PR #3, opened 2019-02-20
Limit the number of repetitions in bir_wp_init_wps_bool_sound_thm.

totorigolo:
I used `bir_wp_init_wps_bool_sound_thm` with a trivial postcondition on a one-block program, and it ran into an infinite loop. The reason: it endlessly tried to prove `F`.

`20` is totally arbitrary here, I don't know if this value should be higher or if `TRY` is enough. However, I successfully ran both `./toy1-test.exe` and `./aes-test.exe` (stopped at `"remaining: 20"`) with just `TRY`.

### DIFF
--- a/src/tools/wp/bir_wpLib.sml
+++ b/src/tools/wp/bir_wpLib.sml
@@ -30,9 +30,12 @@ struct
 (* establish wps_bool_sound_thm for an initial analysis context (program, post, ls, wps) *)
 fun bir_wp_init_wps_bool_sound_thm (program, post, ls) wps defs =
       let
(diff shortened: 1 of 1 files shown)


andreaslindner:
I guess you could make it fail if the goal is F instead. But I merge the pull request as it is since this branch is experimental anyways.

totorigolo:
My specific case was a type error (`Imm64` with 32-bit variable), but I think that more complex situations can happen where the goal isn't `F` but doesn't succeed either.

andreaslindner:
I see. The assumption there was that the goal is right and it's simple to prove :).

PR #6, opened 2019-03-19
Extension of WP tool to the Assume statement

didriklundberg:
Previously, the weakest precondition soundness theorem for `Assume ex` and the postcondition `Q` (`bir_triple_exec_stmtB_assume_thm`) stated `ex /\ Q` as the weakest precondition, which we know to not correspond with the typical semantics of the Assume statement in literature. The Hoare triple (through `bir_pre_post_def`, `bir_exec_block_jmp_triple_def`,  `bir_exec_block_cjmp_triple_def` and `bir_exec_to_labels_triple_def`) has now been changed to take into account the correct semantics of Assume. The rest of the theorems in `bir_wpScript.sml` has been modified accordingly.

In addition to this, I have made some formatting changes to better harmonise with the rest of HolBA.

I also updated `toy1-test.sml`, and made passified versions of the AES test as well as the toy test.

Still on the TODO list is making some of the proofs clearer and more maintainable through lemmatization, taking a look at the proof procedures, and making a tool to obtain the old Hoare triple from the new.


andreaslindner:
Thank you for your work!

PR #10, opened 2019-03-25
Dev pass

didriklundberg:
I suggest the addition of a small, unverified tool for obtaining BIR programs in passified SSA form. Maybe this can be of use to someone.

### DIFF
--- a/src/tools/Holmakefile.gen
+++ b/src/tools/Holmakefile.gen
@@ -1,6 +1,6 @@
 # includes
 # ----------------------------------
-INCLUDES   = ./cfg ./lifter ./wp ./exec
+INCLUDES   = ./cfg ./lifter ./wp ./exec ./pass
 
 
 # configuration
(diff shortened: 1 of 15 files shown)


didriklundberg:
> But I would like if you could move the syntax functions next to the theories with the corresponding definitions. If there is no syntax library yet, we should create it.

Yes, sorry about this. My previous routine was to leave HolBA pristine and untouched and make my own little syntax files for every context. The idea was always to eventually merge everything in a more sensible way, I just forgot to do it in this pull request. 

> Maybe we could also find a better place for the loading and storing of bir programs. What about the new libs directory once Thomas inaugurated it with his additions?

I agree 100%, same for the variable comparison (don't want multiple of these floating around).

andreaslindner:
No problem, I just thought that otherwise I would have used the comment function to feed back before merge. But this really makes no difference in our little project right now. It's just to get used to the modern teamwork functions here. :)

PR #14, opened 2019-03-26
New bir_exp_to_wordsLib 

totorigolo:
New (non proof-producing) library that turns BIR expressions into non-BIR wordsSyntax+combinSyntax expressions.

In the current state, birExpToWordsLib depends on `tools/wp/bir_wp_simpLib` in order to handle "BIR implications" (`bir_exp_imp`) that are defined there. I see 2 solutions to this problem:
- move `bir_exp_imp_def`, `bir_exp_...._def` to somewhere else, like `theories/`
- remove the handling of implications in `birExpToWordsLib`, and add a simplifier somewhere in `wp/` that removes the non-BIR expressions. The downside here is that we will then lose implications in word expressions and when using Z3.

What do you prefer? @andreaslindner


totorigolo:
Everything is working :tada:

PR #20, opened 2019-03-28
WP: wps_id_prefix + BL_Label support + dependencies fix +  MemEq&imp in bir_expLib

totorigolo:
`src/libs/bir_expLib.sml` line 4 is the reason why I had to "fix" a lot of `open`s.

Also, I strangely had to revert this: https://github.com/kth-step/HolBA/commit/3f1b4ae5ae996126bc48b44438304dc4477c04a3 (in 199e398) in order to get `tools/wp/examples/*` working.

Examples in `tools/wp/examples` and `examples/aes` are all working.

### DIFF
--- a/src/libs/bir_expLib.sml
+++ b/src/libs/bir_expLib.sml
@@ -106,6 +106,13 @@ struct
           ((xf "(") cf (ef exp1) cf (xf " ") cf (xf bpredopstr) cf (xf " ") cf (ef exp2) cf (xf ")"))
         end
 
+      else if is_BExp_MemEq exp then
+        let
+          val (exp1, exp2) = (dest_BExp_MemEq) exp;
+        in
+          ((xf "(") cf (ef exp1) cf (xf " = ") cf (ef exp2) cf (xf ")"))
+        end
+
       else if is_BExp_IfThenElse exp then
         let
           val (expc, expt, expf) = (dest_BExp_IfThenElse) exp;
--- a/src/libs/bir_expLib.sml
+++ b/src/libs/bir_expLib.sml
@@ -1,92 +1,66 @@
 structure bir_expLib =
 struct
 
+  local
+
   open HolKernel boolLib liteLib simpLib Parse bossLib;
 
   open bir_expSyntax bir_immSyntax bir_envSyntax bir_exp_immSyntax bir_exp_memSyntax;
+  open bir_bool_expSyntax;
 
+  val ERR = Feedback.mk_HOL_ERR "bir_expLib";
 
+  in
 
+  fun castt_to_string castt =
+    if is_BIExp_LowCast castt then "CL"
+    else if is_BIExp_HighCast castt then "CH"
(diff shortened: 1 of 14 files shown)


totorigolo:
Looks like it broke something in wp_simp

totorigolo:
Fixed :)

totorigolo:
Please, don't build IDs (like "bir_wp_comp_wps_iter_step2_wp_entry") by-hand, I added a function in `bir_wpLib` to do that.

src/libs/bir_expLib.sml
```diff
@@ -156,8 +137,15 @@ struct
           ((xf "(") cf (ef expm) cf (xf ":") cf (xf endistr) cf (xf "[") cf (ef expad) cf (xf "] = ") cf (ef expv) cf (xf ")"))
         end
 
+      else if is_bir_exp_imp exp then
```
andreaslindner: this could be a problem because this whole function assumes "normalized" terms which only contain the basic BIR constructors from core

src/libs/bir_expLib.sml
```diff
@@ -1,92 +1,66 @@
 structure bir_expLib =
 struct
 
+  local
```
andreaslindner: what is the advantage of this local?

src/libs/bir_expLib.sml
totorigolo: I didn't see that in the documentation :stuck_out_tongue: Sure we could remove that, but then we should provide a function to "normalize" a BIR expression. But I don't really see the point of restricting it :slightly_smiling_face:

src/libs/bir_expLib.sml
totorigolo: That's the pattern I'm using in all the files:
```
structure someLib =
struct

  local

  open HolKernel Parse boolLib bossLib;
  more opens;

  val ERR = Feedback.mk_HOL_ERR "someLib";
  val wrap_exn = Feedback.wrap_exn "someLib";

  in

  exported stuff

  end (* local *)

end (* someLib *)
```
Like this, you don't export too much stuff. (do you really want to export `ERR`?)

totorigolo:
I replied to your comments :slightly_smiling_face:

src/libs/bir_expLib.sml
andreaslindner: the problem is that this kind of function always has some scope and i find the basic set of expression types a natural fit unless we have another one. but i would leave it like this for now also. it needs to be revisited anyways I think. and possibly we have other infrastructure around at that time to describe a clear scope or similar.

src/libs/bir_expLib.sml
andreaslindner: good point. so in this way you manage this ad-hoc interface creation. in the end of the day, we should get used to a cleaner practice and create interfaces I guess, but you're absolutely right in this. good that I asked :)

PR #33, opened 2019-04-22
Changing memory representation from a function to a finite map (and add BExp_MemConst)

andreaslindner:
This change should simplify working with execution tools, SMT solver and others as well. In fact, the memory semantics was already a finite mapping before. After this change it is just more explicit and I believe that this eases our proofs.

Please tell me what you think in general and also about the concrete changes.

PS: I also adjusted the HolSmtLib code in our HOL to work with the finite maps instead of functions. Please have a look at this as well.
[HOL for HolBA with fmap HolSmtLib](/kth-step/HOL/tree/for_holba_smtfmap)

### DIFF
--- a/src/libs/bir_exp_to_wordsLib.sml
+++ b/src/libs/bir_exp_to_wordsLib.sml
@@ -411,7 +411,7 @@ struct
           val writes = offset_writes_up_to nsplits []
             handle e => raise wrap_exn "bir_exp_to_words::store::writes" e;
           (* Fold using mk_comb *)
-          val whole_store_tm = List.foldl (fn (update_tm, mem_tm) =>
+          val whole_store_tm = List.foldr (fn (update_tm, mem_tm) =>
             mk_comb (update_tm, mem_tm))
             mem_w writes
             handle e => raise wrap_exn "bir_exp_to_words::store::mk_comb" e;
--- a/src/libs/bir_exp_to_wordsLib.sml
+++ b/src/libs/bir_exp_to_wordsLib.sml
@@ -170,7 +170,7 @@ struct
                 val val_ty = word_ty_of_bir_immtype_t val_bir_ty
                   handle e => raise wrap_exn "bir_exp_to_words::den::mem" e
               in
-                Type `: ^addr_ty -> ^val_ty`
+                Type `: ^addr_ty |-> ^val_ty`
               end
             else
               raise Fail ("unhandled type: " ^ (term_to_string bir_type))
@@ -298,7 +298,7 @@ struct
         val w = bir_exp_to_words exp;
         *)
         (*
-         * ((addr+0) :> mem) @@ ((addr+1) :> mem) @@ ...
+         * ((addr+0) ' mem) @@ ((addr+1) ' mem) @@ ...
          *)
         let
           val (bir_mem, bir_addr, bir_endi, bir_val_ty) = dest_BExp_Load exp
@@ -321,7 +321,8 @@ struct
                 val offset = Arbnumcore.- (n, Arbnumcore.one);
                 val offset_w = mk_word (offset, size_of base_addr_w);
                 val addr = ``^base_addr_w + ^offset_w``;
-                val load_tm = ``^addr :> ^mem_w``;
+                (* this may need to be refined to: case FLOOKUP .. .. of SOME x => x | NONE => 0w *)
+                val load_tm = ``FAPPLY ^mem_w ^addr``;
               in
                 offset_reads_up_to (Arbnumcore.- (n, Arbnumcore.one)) (load_tm::acc)
               end
@@ -386,6 +387,10 @@ struct
           val write_len_num = Arbnumcore.fromInt write_len
           val write_len_ty = mk_numeric_type write_len_num
           val val_w = bir_exp_to_words bir_val
+          (*
+          val n = nsplits;
+          val acc = [];
+          *)
           fun offset_writes_up_to n acc =
             if n = Arbnumcore.zero then acc else
               let
@@ -399,7 +404,7 @@ struct
                 val high_bit_tm = mk_numeral high_bit
                 val val_tm = mk_word_extract
                   (high_bit_tm, low_bit_tm, val_w, write_len_ty)
-                val store_tm = ``(^addr_tm =+ ^val_tm)``
+                val store_tm = ``(\x. (FUPDATE x (^addr_tm, ^val_tm)))``
               in
                 offset_writes_up_to (Arbnumcore.- (n, Arbnumcore.one)) (store_tm::acc)
               end
@@ -411,7 +416,7 @@ struct
             mem_w writes
             handle e => raise wrap_exn "bir_exp_to_words::store::mk_comb" e;
         in
-          whole_store_tm
+          (snd o dest_eq o concl o (SIMP_CONV pure_ss [boolTheory.BETA_THM])) whole_store_tm
         end
       (*** WP specific terms ***)
       (* Implications *)
--- a/src/libs/bir_exp_to_wordsLib.sml
+++ b/src/libs/bir_exp_to_wordsLib.sml
@@ -148,7 +148,7 @@ struct
           handle e => raise wrap_exn "bir_exp_to_words::const" e
       (* Memory constants *)
       else if is_BExp_MemConst exp then
-        raise Fail ("unhandled: BExp_MemConst")
+        raise ERR "bir_exp_to_words" "unhandled: BExp_MemConst"
       (* Den *)
       else if is_BExp_Den exp then
         (* Manual tests
--- a/src/libs/bir_exp_to_wordsLib.sml
+++ b/src/libs/bir_exp_to_wordsLib.sml
@@ -146,6 +146,9 @@ struct
       if is_BExp_Const exp then
         (snd o gen_dest_Imm o dest_BExp_Const) exp
           handle e => raise wrap_exn "bir_exp_to_words::const" e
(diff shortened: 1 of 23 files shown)


didriklundberg:
This looks good. I don't immediately see the practical usage of changing the memory representation but I also can't see how it makes anything worse, so I am sure you're right in making this change.

src/tools/exec/bir_exec_expLib.sml
```diff
@@ -181,74 +174,54 @@ val UPDATE_EQ_num = prove(``
   SIMP_TAC std_ss [UPDATE_EQ]
 );
 
+fun combin_purge_conv ct =
```
totorigolo: This isn't directly related to this PR, but can you add a comment explaining what this functions does (as it is a public function of this lib)? Ideally, write a `.sig` file with short comments :slightly_smiling_face:

Please read the next comment line 221. Then, if you want to keep this function, you can just write a comment in this file as I just created a new issue for this: #34

src/tools/exec/bir_exec_expLib.sml
```diff
@@ -181,74 +174,54 @@ val UPDATE_EQ_num = prove(``
...
-                  (fn x => REWRITE_CONV [Once bir_update_mmap_def] x) THENC
-                  (fn x => if is_bir_update_mmap x then
-                             RATOR_CONV (LAND_CONV (EVAL THENC combin_purge_conv)) x
-                           else
-                             raise UNCHANGED)
+            (fn x => REWRITE_CONV [Once bir_update_mmap_def] x) THENC
+	    (fn x => if is_bir_update_mmap x then
+		       RATOR_CONV (LAND_CONV (
+			 EVAL THENC
+			 (fn x => REWRITE_CONV [Once finite_mapTheory.FUPDATE_PURGE] x) THENC
```
totorigolo: This was the only usage of `combin_purge_conv`, so I wonder if `combin_purge_conv` is still needed? (i.e. maybe you should remove it entirely)

src/tools/lifter/bir_exp_liftingScript.sml
```diff
@@ -7,9 +7,305 @@ open bir_exp_memTheory bir_bool_expTheory
...
+
+val bir_mf2mm_def = Define `
+   (bir_mf2mm (mf: 'a word -> 'b word)) : 'a word |-> 'b word =
+      FUN_FMAP mf (UNIV:'a word -> bool)(*{ x | mf x <> 0w }*)
+    `;
+
+
+
+val bir_FLOOKUP_mmap_w2n__thm = store_thm("bir_FLOOKUP_mmap_w2n__thm", ``
+  !mmap_w a. FLOOKUP (bir_mmap_w_w2n mmap_w) (a MOD dimword (:'a)) = OPTION_MAP w2n (FLOOKUP mmap_w (n2w a : 'a word))
```
totorigolo: Can you break these long lines? :grin: (this one and all the long definitions below)

src/tools/lifter/bir_lifting_machinesScript.sml
```diff
@@ -171,7 +171,9 @@ SIMP_TAC (std_ss++bir_TYPES_ss) [bir_machine_lifted_mem_def,
   bir_env_vars_are_initialised_EMPTY,
   bir_env_var_is_initialised_def,
   bir_env_read_NEQ_UNKNOWN, type_of_bir_val_def,
-  GSYM LEFT_FORALL_IMP_THM]);
+  GSYM LEFT_FORALL_IMP_THM] >>
+REWRITE_TAC [n2w_bir_load_mmap_w2n_thm]
```
totorigolo:
```suggestion
  REWRITE_TAC [n2w_bir_load_mmap_w2n_thm]
```

src/libs/bir_exp_to_wordsLib.sml
```diff
@@ -321,7 +321,8 @@ struct
                 val offset = Arbnumcore.- (n, Arbnumcore.one);
                 val offset_w = mk_word (offset, size_of base_addr_w);
                 val addr = ``^base_addr_w + ^offset_w``;
-                val load_tm = ``^addr :> ^mem_w``;
+                (* this may need to be refined to: case FLOOKUP .. .. of SOME x => x | NONE => 0w *)
```
totorigolo: Since you wrote this comment, I wonder why you didn't do it. The semantic is that the value is `0w` if the address isn't mapped, so only using `FAPPLY` doesn't respect the semantic? Don't we risk soundness issue with the current code?

src/libs/examples/z3-sat-model-test.sml
```diff
@@ -41,7 +43,11 @@ fun produce_sat_thm term model =
 
 (* Get a SAT model using Z3 (this function assumes that the given term is SAT) *)
 val term = ``(z + y = 2 * x) /\ ((x * x + y - 25) = z:int)``;
+(* (FUN_MAP2 (K 0w) (UNIV)) *)
+(*
+val term = ``(m2 = FUPDATE (m1: word32 |-> word8) (3w, 2w)) /\ (FAPPLY m2 3w = 2w)``;
```
totorigolo: Can you move this commented-out term below as not-commented code? This will enable using `xxx-test.sml` examples in the CI (maybe they already are, I don't know).

src/libs/bslSyntax.sml
```diff
@@ -218,6 +218,21 @@ struct
...
+  (* Memory constants (BExp_MemConst: bir_exp_t) *)
+  fun bconstmem_helper [] =
+        finite_mapSyntax.mk_fempty (numSyntax.num, numSyntax.num)
+    | bconstmem_helper ((k,v)::l) =
+        finite_mapSyntax.mk_fupdate (bconstmem_helper l,
+          pairSyntax.mk_pair (numSyntax.term_of_int k, numSyntax.term_of_int v));
+
+  fun bconstmem al vl l =
+    let
+      val aty = Term [QUOTE ("Bit" ^ (Int.toString al))];
```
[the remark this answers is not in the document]
didriklundberg: Yes, this is better.

andreaslindner:
> src/libs/bir_exp_to_wordsLib.sml  line 324: I wonder if the current code is correct.

If you refer to the comment then that would be actually correct according to the semantics we have right now. The FAPPLY I wrote is only correct if the finite map covers the whole address range. Since we cannot deal with FLOOKUP right now this is the best we can do though. Bottomline: we simply assume that the whole address range is covered and this is no problem, because such a finite map can easily be derived from the partial one.

andreaslindner:
> did you check that all the code using HolSmtLib still works? This doesn't seem to be checked by the CI yet. I can check if you want. Alternatively, we can work on #32 first by adding example/xxx-test.sml files to the CI to solve this question. This depends if you want to merge this now or if you can wait some time.

I played with it a bit and I think we should check this with the CI, as in your alternative suggesstion. I would be happy if you could write test cases after we have completed the testing pull request (#32 ) and then have rebased this stuff here.

src/libs/bslSyntax.sml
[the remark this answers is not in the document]
andreaslindner: I would like it if they would generate a type but they generate Imms. I looked for one and didn't find one that generates the type. I suppose we will have to write this Syntax function and then we can use it here :).

src/tools/exec/bir_exec_expLib.sml
andreaslindner: I would like to take this separately to resolve the issue you created for the hole repository or stepwise big parts of it.

src/tools/exec/bir_exec_expLib.sml
andreaslindner: Yes, it is not needed anymore after the memory is changed to a finite map. I forgot to remove it and will do it now.

src/libs/bir_exp_to_wordsLib.sml
andreaslindner: FAPPLY is undefined for unmapped values. This function is for SMT solver interaction and we just assume fully defined maps for now.

src/libs/examples/z3-sat-model-test.sml
andreaslindner: I think we should do this exactly when we write the test.sml file

src/libs/bir_exp_to_wordsLib.sml
totorigolo: > FAPPLY is undefined for unmapped values

Oh right, so it's sound, but not complete. I am right?

src/libs/bslSyntax.sml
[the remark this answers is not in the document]
andreaslindner: It's already there, of course. This is the one: `bir_immtype_t_of_size`.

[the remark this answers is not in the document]
andreaslindner: But nobody saw the copy-paste error in the next line 😏

src/libs/bir_exp_to_wordsLib.sml
andreaslindner: Under the assumption that we are dealing with fully defined maps it should be both if I don't miss anything now. In this case our semantics is equal to FAPPLY.

totorigolo: Yes, under this assumption I think you're right. I was saying without this assumption.

I will mark this as resolved now that we are aware of this :)

totorigolo:
I marked approved, but maybe we should still wait on #32?

andreaslindner:
Yes, we should work out the tests and I'm again thinking about the implications of these changes. Just to be sure. No need to rush.

PR #35, opened 2019-05-02
Travis: Slack notifications and encrypt email

totorigolo:
I encrypted Andreas' email address to prevent forks to send him emails about Travis status.


andreaslindner:
Could you remove my mail address completely? I read that travis would send mails to comitter and author in this case.

andreaslindner:
You want to squash? or keep both?

PR #38, opened 2019-05-03
Update the dependency diagram

totorigolo:


andreaslindner:
Why "may be wrong"?

totorigolo:
Because it looks like a bunch of spaghetti :sweat_smile:  I just wanted you to double check :)

andreaslindner:
It is correct but I think we need to improve it. There are some hidden dependencies also. What about the following?

- the theory directory should be one dependency blob and shared another one and tools another one. this is because we know that there are dependencies in many directions there and this diagram is supposed to give an overview i would say. we could have more detailed ones at other places maybe

- the green arrows indicating the holheap build order could be more pale. this is less often interesting and it's just an order. alternatively it can also be encoded with numbers behind the directory names. in the latter solution it would not influence the layout

- the examples may be not required in this diagram and can be just removed for clarity

totorigolo:
> It is correct but I think we need to improve it. There are some hidden dependencies also. What about the following?
> * the theory directory should be one dependency blob and shared another one and tools another one. this is because we know that there are dependencies in many directions there and this diagram is supposed to give an overview i would say. we could have more detailed ones at other places maybe

You mean something like this?

```
[theory] <-- [shared] <-- [tools]

<-- meaning "depends on"
```

We don't need to use GraphViz then :stuck_out_tongue: 

> * the green arrows indicating the holheap build order could be more pale. this is less often interesting and it's just an order. alternatively it can also be encoded with numbers behind the directory names. in the latter solution it would not influence the layout

Or we could also just remove it entirely and write it as text in the README. (they don't influence the layout though)

> * the examples may be not required in this diagram and can be just removed for clarity

Let me try something.

andreaslindner:
I meant that the blob contains the dependencies within. like we talked about the sml module dependencies once, you remember? and outside you only have the dependencies between the blobs, but nothing goes into the blobs and is by this accumulated so to say.

since it is generated, i would prefer the numbering over putting it in some other more static place.

totorigolo:
> I meant that the blob contains the dependencies within. like we talked about the sml module dependencies once, you remember? and outside you only have the dependencies between the blobs, but nothing goes into the blobs and is by this accumulated so to say.

Oh yes I remember. But we would lose quite a lot.

> since it is generated, i would prefer the numbering over putting it in some other more static place.

It's not. My script works (for the moment at least) only at the file level, not folder level.

andreaslindner:
Regarding your last sentence - I was referring to what you wrote here about the heap build order, see:

"Or we could also just remove it entirely and write it as text in the README. (they don't influence the layout though)"

andreaslindner:
when opening the branch and look at the rendered readme.md i cannot see your latest png file for some reason. is it the same for you? (the same happens when i browse into doc to the file)

totorigolo:
> when opening the branch and look at the rendered readme.md i cannot see your latest png file for some reason. is it the same for you? (the same happens when i browse into doc to the file)

There reason here is quite funny :sweat_smile: `dot -Tpdf` instead of `dot -Tpng`. That's also the reason of the force push. You just have to empty your cache.

Btw, in case you don't know it, GitHub can show a [rich diff](https://github.com/kth-step/HolBA/pull/38/files?short_path=c659cc4#diff-c659cc4bed1dcccf2c09632f198b9bb2).

andreaslindner:
The rich diff looks good! And also the diagram is now very clear. I guess you apply your transitive analysis there, right? Do you know why scamv is dependent on exec? I don't recall now.

totorigolo:
> > since it is generated, i would prefer the numbering over putting it in some other more static place.
> 
> It's not. My script works (for the moment at least) only at the file level, not folder level.

As I said, this diagram is hand-written, not auto-generated.

> The rich diff looks good! And also the diagram is now very clear. I guess you apply your transitive analysis there, right? Do you know why scamv is dependent on exec? I don't recall now.

I added the `SCAM-V --> exec` dependency just because you told me so at some point :stuck_out_tongue:

andreaslindner:
hahaha, i see. then i forgot why i said so.

to sum up, you want to add the handwritten diagram for now?

andreaslindner:
Of course! Because of the symbolic execution...

In this case. Some changes maybe:
- theories should be theory
- you could try to make two rectangular subgraphs. one for theory and one for tools. label them and then you can drop the prefixes in the current nodes accordingly. and keep all the arrows - you are right dropping them would remove quite a lot. this would make this diagram more organized i think
- add numbers for the holheap build order in parenthesis at the nodes (after the names)

andreaslindner:
wait

andreaslindner:
you maybe want to remove (outdated) in the README.md

PR #39, opened 2019-05-03
Some libs and examples

totorigolo:
Soooooo, this PR is quite big. I know, that's bad. That's some of my local changes I had before the directoryquake. But this should be quite easy to review since every commit should be atomic.

**However**, this PR requires Z3 in order to pass the CI (@andreaslindner, what's the status?).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,8 @@ HOLMAKEFILE_DIRS = $(patsubst %/,%,$(sort $(foreach file,$(HOLMAKEFILE_GENS),$(d
 SML_RUNS         = $(foreach sml,$(call rwildcard,$(SRCDIR)/,*.sml),$(sml)_run)
 SML_RUNQS        = $(foreach sml,$(call rwildcard,$(SRCDIR)/,*.sml),$(sml)_runq)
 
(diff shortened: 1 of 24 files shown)


andreaslindner:
let's do it like this: we add it in the CI setup phase and add a simple test in one comment and a separate PR. if this works, we rebase it into master and rebase this one on top of it and let the CI deal with the problem :)

andreaslindner:
could you try the first step or do you need my help for this? you just need to add another line to the before_script parts of the two prepare jobs in the travis file. this should be the install_z3 script. the rest should work magically if i didn't leave something for us to debug in these scripts 😄

totorigolo:
Rebased on master.

totorigolo:
The commit order seems wrong on GitHub, because they are not in chronological order (because :cherries:-pick and rebase magic). Keep that in mind when reviewing :slightly_smiling_face:

andreaslindner:
So if I use some git GUI tools it would probably be correct since I'm looking at the actual commit order not timestamps, right?

totorigolo:
Yep, I think so. If you have any doubt, check with `git log` :slightly_smiling_face:

PR #43, opened 2019-05-06
scripts: drop sh support + improve find HolBA in autoenv

totorigolo:
- `scripts/: fix for ZSH (use [[ ]] instead of [ ])`
  * Doesn't work with `sh`, but it's simpler and safer: https://serverfault.com/a/52050

- `autoenv.sh: better find HolBA dir`
  * `source ./scripts/setup/autoenv.sh` failed when executed in HolBA root directory.


totorigolo:
I closed the PR to reduce Travis' work. It's passing now, but I'm not convinced with the code.

PR #45, opened 2019-05-07
Scripts: ZSH support

totorigolo:
=> `scripts/: fix for ZSH (use [[ ]] instead of [ ])`

Doesn't work with `sh`, but it's simpler and safer: https://serverfault.com/a/52050
The initial goal was to fix this: https://unix.stackexchange.com/questions/255480/why-does-behave-differently-inside-in-zsh-and-bash

=> `autoenv.sh: fix sourcing with ZSH`

`BASH_SOURCE` doesn't exist in ZSH. This is a problem when sourcing `autoenv.sh` in ZSH. Fix as suggested [here](https://stackoverflow.com/a/54755784).

----
Follows #43, since it cannot be reopened because of force-push.

### DIFF
--- a/scripts/setup/autoenv.sh
+++ b/scripts/setup/autoenv.sh
@@ -3,7 +3,7 @@
 OPT_DIR_PARAM=$1
 
 # get setup directory path
-SETUP_DIR=$(dirname "${BASH_SOURCE[0]}")
+SETUP_DIR=$(dirname "${BASH_SOURCE[0]:-${(%):-%x}}")
 SETUP_DIR=$(readlink -f "${SETUP_DIR}")
 
 
@@ -27,14 +27,29 @@ fi
 # find HolBA dir for this script
 #####################################################
 
-
-if [ -z "${HOLBA_DIR}" ]; then
+if [[ -z "${HOLBA_DIR}" ]]; then
   HOLBA_DIR=${SETUP_DIR}/../..
   HOLBA_DIR=$(readlink -f "${HOLBA_DIR}")
 fi
 
-if [[ ! -d "${HOLBA_DIR}/scripts/setup" ]]; then
-  echo "ERROR: HOLBA_DIR not found (tried $HOLBA_DIR)"
+function is_holba {
+  # Apparently not a good practice (https://stackoverflow.com/a/43840545)
+  # but this enables to easily "echo then return $FALSE"
+  local FALSE="1"
+
+  [[ -d "$1/scripts/setup" ]] \
+      || { echo "bad scripts dir" > /dev/stderr; return $FALSE; }
+  [[ "$(head -n1 "$1/README.md" 2> /dev/null | grep -c '# HolBA')" -eq 1 ]] \
+      || { echo "bad README.md" > /dev/stderr; return $FALSE; }
+  local REMOTE_REGEX='github[.]com[:/]kth-step/HolBA'
+  [[ "$(cd "$1" && git remote -v | grep -c $REMOTE_REGEX)" -gt 0 ]] \
+      || { echo "missing Git remote" > /dev/stderr; return $FALSE; }
+
+  true
+}
+
+if ! is_holba "${HOLBA_DIR}"; then
+  echo "ERROR: HOLBA_DIR not found or incorrect (tried: '${HOLBA_DIR}')"
   return 2
 fi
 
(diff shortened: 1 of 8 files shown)


scripts/setup/autoenv.sh
```diff
@@ -29,12 +29,12 @@ fi
 
 
 if [ -z "${HOLBA_DIR}" ]; then
-  HOLBA_DIR=${SETUP_DIR}/../..
+  HOLBA_DIR="${SETUP_DIR}/../.."
```
andreaslindner: Do you think the quotes are necessary here? I consistently didn't put any in these places :)

scripts/setup/autoenv.sh
```diff
@@ -29,12 +29,12 @@ fi
...
   HOLBA_DIR=$(readlink -f "${HOLBA_DIR}")
 fi
 
 if [[ ! -d "${HOLBA_DIR}/scripts/setup" ]]; then
```
andreaslindner: What about adding your README.md head, grep, "HolBA" check also?

scripts/setup/autoenv.sh
totorigolo: Ah? Oh. Oops? I don't think that they are necessary.

**However**, I just noticed that this `if` is using single `[ ]`.

andreaslindner: Oh, true, I didn't really look for this anyways, you could grep something like "^[[ " and " ]^]" where ^ should be the not symbol. This to find the cases you missed.

But I meant it like a question, do you think there is an advantage of using quotes or a disadvantage of not using them for paths? I used them everywhere else consistently (I tried to be consistent).

scripts/setup/autoenv.sh
totorigolo: Which one do you prefer?
```bash
head -n1 README.md | grep -c '# HolBA'
```

or

```bash
git remote -v | grep -c 'github.com:kth-step/HolBA'
```

andreaslindner: Hmm, both have advantages and disadvantages. We could add both for now so that we have them and make it robust now. And then we would remove one or adjust it as soon as it is required. What do you think about this cannonball solution?

totorigolo: Did both + the current one, so 3 tests in total :)

scripts/setup/autoenv.sh
totorigolo: I'm not sure, but I would say that it's safer to use quotes everywhere, because there could be a space in `$SETUP_DIR`. Here it shouldn't pose a problem, but you really want to quote in `rm -rf "$SETUP_DIR/../../some/path/"`. Then, if you want to be consistent, you should quote everywhere :slightly_smiling_face:

scripts/setup/autoenv.sh
```diff
@@ -27,14 +27,25 @@ fi
...
+  [[ "$(head -n1 "$1/README.md" 2> /dev/null | grep -c '# HolBA')" -eq 1 ]] \
+      || { echo "bad README.md" > /dev/stderr; return 1; }
+  local REMOTE_REGEX='github[.]com[:/]kth-step/HolBA'
+  [[ "$(cd "$1" && git remote -v | grep -c $REMOTE_REGEX)" -gt 0 ]] \
+      || { echo "missing Git remote" > /dev/stderr; return 1; }
+
+  return 0
+}
+
+if ! is_holba "${HOLBA_DIR}"; then
```
andreaslindner: Just a question. Why not using `[[ ]]` here for consistency?

totorigolo: Because 

```bash
[[ ! $(is_holba "${HOLBA_DIR}") ]]  # not sure if we really need $() here

# is equivalent to:

[[ ! -n $(is_holba "${HOLBA_DIR}") ]] # result is always false
# or something like this, I'm not sure:
[[ -n ! $(is_holba "${HOLBA_DIR}") ]]
```
basically, all non-empty strings will be accepted. That's not what we want.

> `man test`:
>
>STRING equivalent to -n STRING

So we would need this: (well, not exactly, this doesn't seem to work)

```bash
[[ ( ! $(is_holba "${HOLBA_DIR}") ) ]]
```

This is quite lengthy and less readable, and I'm not sure about what it does ^^ I guess that I should learn Bash :sweat_smile: 

---

This syntax comes from this SO answer: https://stackoverflow.com/a/5431932

But reading [this one](https://stackoverflow.com/a/5431932) tells that `return 0 # true` and `return 1 # false` should be avoided. Argh Bash...

Issue #44, opened 2019-05-06
Typo in lifter theorem?


andreaslindner:
Should this be `bs'` instead?

https://github.com/kth-step/HolBA/blob/b49db276f12896eb3bb2f0049bdb2a969f2bf2de/src/theory/tools/lifter/bir_inst_liftingScript.sml#L2008

didriklundberg:
Yes, this looks like a typo for sure. At the very least it is nonsensical since the conclusion resulting from the Running case is already an assumption in the theorem.

PR #47, opened 2019-05-07
Fixed typo in bir_inst_liftingScript.sml

didriklundberg:
I fixed the typo in `bir_inst_liftingScript.sml` (closes https://github.com/kth-step/HolBA/issues/44) and adjusted the proof accordingly. Nothing else seems to be affected.

### DIFF
--- a/src/theory/tools/lifter/bir_inst_liftingScript.sml
+++ b/src/theory/tools/lifter/bir_inst_liftingScript.sml
@@ -2005,7 +2005,7 @@ val bir_is_lifted_prog_MULTI_STEP_EXEC = store_thm ("bir_is_lifted_prog_MULTI_ST
             (case bs'.bst_status of
               | BST_Running =>
                   (?li. MEM (BL_Address li) (bir_labels_of_program p) /\
-                        (bs.bst_pc = bir_block_pc (BL_Address li)))
+                        (bs'.bst_pc = bir_block_pc (BL_Address li)))
               | BST_JumpOutside ll =>
                   ~(MEM ll (bir_labels_of_program p)) /\
                   (IS_BL_Address ll)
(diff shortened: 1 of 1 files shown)


totorigolo:
**Quick tip**: to enable GitHub magic, write `Closes #44` in the commit message or in the PR description. That way, GitHub will automatically close the related Issue when this PR will be merged. I think that you can just replace `mentioned in #44` by `closes #44` and that should do it.

andreaslindner:
I changed Didrik's comment :D. let's see whether it works when we merge it later.

didriklundberg:
> **Quick tip**: to enable GitHub magic, write `Closes #44` in the commit message or in the PR description. That way, GitHub will automatically close the related Issue when this PR will be merged. I think that you can just replace `mentioned in #44` by `closes #44` and that should do it.

Yeah, I usually use the "resolve" keyword. Don't know why I forgot it this time.

didriklundberg:
> Let's wait for the CI to accept and then "rebase and merge" it in.

Hmm.. While Poly/ML 5.8 succeeds in compiling, 5.7 times out seemingly randomly when compiling `bir_exp_tautologiesScript.sml` to theory. However, I managed to compile everything on my own computer using 5.7.

What do you say, @andreaslindner? If this instability of 5.7 is consistent I don't really see a reason to continue using it.

andreaslindner:
This is surely only a timeout like we discussed in slack. You can either just "rebase and merge" or restart the failing job and wait for it to succeed and then click the merge button. Do as you like, both is fine with me.

But no, the instability is not consistent. Sometimes it is poly 5.7.1 and sometimes it is 5.8. I think since I added 5.8 it was even hanging more than 5.7.1 - my guess is that this is just coincidence though, and that the problem is the same in both of them. But we can drop 5.7.1 I think. The only disadvantage I can see is that the default seems to still be 5.7.1 even in the latest commit of HOL (while we "only" use the latest release version): https://github.com/HOL-Theorem-Prover/HOL/blob/master/developers/install-poly.sh. But this shouldn't matter so much. What do you think?

didriklundberg:
> This is surely only a timeout like we discussed in slack.

Yes, that seems to be the case.

> But no, the instability is not consistent. Sometimes it is poly 5.7.1 and sometimes it is 5.8. I think since I added 5.8 it was even hanging more than 5.7.1 - my guess is that this is just coincidence though, and that the problem is the same in both of them. But we can drop 5.7.1 I think. The only disadvantage I can see is that the default seems to still be 5.7.1 even in the latest commit of HOL (while we "only" use the latest release version): https://github.com/HOL-Theorem-Prover/HOL/blob/master/developers/install-poly.sh. But this shouldn't matter so much. What do you think?

HOL has probably stayed at 5.7.1 just because this is the latest package for Ubuntu. I don't think upgrading is a problem as long as we make sure everything runs fine, but on the other side, if the timeouts are indeed evenly distributed across 5.7.1 and 5.8 I see no reason to upgrade at this time.

PR #48, opened 2019-05-08
Some fixes for non-'a observation types (+ 1 PP bugfix + 2 wrap_exn)

totorigolo:
The commit messages should be clear enough :slightly_smiling_face:

### DIFF
--- a/src/shared/bslSyntax.sml
+++ b/src/shared/bslSyntax.sml
@@ -199,7 +199,8 @@ struct
     let
       (* Instantiate the observation type for all statements *)
       val bl_list_obs_ty = List.map
-        (fn (a, l_stmts, b) => (a, List.map (inst [alpha |-> obs_ty]) l_stmts, b))
+        (fn (a, l_stmts, b) => (a,
+          List.map (inst [mk_bir_program_t_ty alpha |-> mk_bir_program_t_ty obs_ty]) l_stmts, b))
         bl_list
       (* list of terms to term of list *)
       val list_tm = List.map
--- a/src/shared/bslSyntax.sml
+++ b/src/shared/bslSyntax.sml
@@ -196,8 +196,18 @@ struct
   val bprog = mk_BirProgram
     handle e => raise wrap_exn "bprog" e
   fun bprog_list obs_ty bl_list =
-    (curry2 mk_BirProgram_list)
-      obs_ty (List.map (uncurry3 ((curry4 mk_bir_block_list) obs_ty)) bl_list)
+    let
+      (* Instantiate the observation type for all statements *)
+      val bl_list_obs_ty = List.map
+        (fn (a, l_stmts, b) => (a, List.map (inst [alpha |-> obs_ty]) l_stmts, b))
+        bl_list
+      (* list of terms to term of list *)
+      val list_tm = List.map
+        (uncurry3 ((curry4 mk_bir_block_list) obs_ty))
+        bl_list_obs_ty
+    in
+      mk_BirProgram_list (obs_ty, list_tm)
+  end
     handle e => raise wrap_exn "bprog_list" e
 
   (****************************************************************************)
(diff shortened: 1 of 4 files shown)


andreaslindner:
> That's strange, the Travis CI - Branch build had finished before I even created this PR, yet it still appears as "In progress". Anyways, all checks are successful :)

Could it be that you pushed to the branch a while before you created the pull request? These two things should be independent as I understand.

totorigolo:
Yes, that's exactly what I did (on purpose). I pushed, waited to see if the "CI - Branch" passes, and then created the PR.

andreaslindner:
> Yes, that's exactly what I did (on purpose). I pushed, waited to see if the "CI - Branch" passes, and then created the PR.

Now I got it. I think I got confused and focused on that it finished before creation. Whatever  - now I see your point.

totorigolo:
Anyway, we can just restart the build, we're not in a hurry :)

andreaslindner:
I think there is no need even ;). But we could see whether it would update the state. I'll press restart now to see later

src/shared/bslSyntax.sml
```diff
@@ -196,8 +196,18 @@ struct
   val bprog = mk_BirProgram
     handle e => raise wrap_exn "bprog" e
   fun bprog_list obs_ty bl_list =
-    (curry2 mk_BirProgram_list)
-      obs_ty (List.map (uncurry3 ((curry4 mk_bir_block_list) obs_ty)) bl_list)
+    let
+      (* Instantiate the observation type for all statements *)
+      val bl_list_obs_ty = List.map
+        (fn (a, l_stmts, b) => (a, List.map (inst [alpha |-> obs_ty]) l_stmts, b))
```
andreaslindner: could it happen that the l_stmts already have some terms (statements) which don't have alpha as an observation type? for example when there is an observe statement?

src/tools/wp/bir_wpLib.sml
```diff
@@ -255,25 +258,24 @@ struct
...
-        !bir_wp_comp_wps_iter_step2_consts;
+      (* FIXME: This seems to take some time, is that normal? *)
+      val wps_eval_restrict_consts = !bir_wp_comp_wps_iter_step2_consts;
+      val obs_ty = (hd o snd o dest_type o type_of) program
       val wps1_thm =
         computeLib.RESTR_EVAL_CONV wps_eval_restrict_consts
           (list_mk_comb
             (* TODO: Add to bir_wpSyntax *)
-            (``bir_wp_exec_of_block:'a bir_program_t ->
+            (inst [Type `:'obs_ty` |-> obs_ty]
```
andreaslindner: I cannot test this right now. Could you explain to me briefly why this makes sense?

src/shared/bslSyntax.sml
totorigolo: I haven't tried, so I should write tests, but if there is no alpha, then `inst` should have no effect.

src/tools/wp/bir_wpLib.sml
totorigolo: If the observation type in `program` isn't `'a`, then `list_mk_comb` will fail because of type mismatch. So I'm instantiating the observation type to the program's obs type.

src/shared/bslSyntax.sml
andreaslindner: or alpha means something else possibly

totorigolo: Oh, right. We should match on `'a bir_program_t` instead.

src/tools/wp/bir_wpLib.sml
andreaslindner: oh i see, that makes a lot of sense

src/shared/bslSyntax.sml
andreaslindner: yes, if `inst` is a pattern match, then this would make it stable

andreaslindner: are you going to implement the match on `'a bir_program_t`?

totorigolo: @andreaslindner Did the new commit fix this?

totorigolo:
The last blocker from the review should be fixed now.

PR #49, opened 2019-05-10
Added more conditional rewriting theorems in bir_bool_equivScript.sml

didriklundberg:
What the title says. This should be useful when reducing BIR expressions to something that an SMT solver can chew on.

### DIFF
--- a/src/theory/bir-support/bir_exp_equivScript.sml
+++ b/src/theory/bir-support/bir_exp_equivScript.sml
@@ -1,7 +1,8 @@
 open HolKernel Parse boolLib bossLib;
 
 (* From /core: *)
-open bir_immTheory bir_expTheory bir_exp_immTheory;
+open bir_immTheory bir_expTheory bir_exp_immTheory
+     bir_typing_expTheory bir_valuesTheory;
 
 (* From /theories: *)
 open bir_bool_expTheory;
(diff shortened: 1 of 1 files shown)


totorigolo:
Completely unrelated to the actual changes, but can you rebase you branch to avoid having a merge commit?

I'm looking into the code now :slightly_smiling_face:

didriklundberg:
> Completely unrelated to the actual changes, but can you rebase you branch to avoid having a merge commit?

Ok, I used some Jedi powers that I think solved the issue.

totorigolo:
Wait, why do you need the hypothesis in `bir_disj_equiv`? `bir_and_equiv` is almost the same and doesn't have them @didriklundberg

didriklundberg:
> Wait, why do you need the hypothesis in `bir_disj_equiv`? `bir_and_equiv` is almost the same and doesn't have them @didriklundberg

It follows directly from the semantics of BIR conjunction that both conjuncts are Boolean. This is not the case for BIR disjunction, in which only one of the disjuncts is required to be Boolean for the disjunction to hold. Accordingly, to map onto HOL disjunction we must require that both disjuncts be Boolean.

totorigolo:
That makes sense, thanks :)

PR #50, opened 2019-05-13
Add simplification library in bir-support

andreaslindner:
Now we can simplify to get a set of variables of a program by using:

```
open HolBASimps;
val prog = ``BirProgram ...``
val t = SIMP_CONV (pure_ss++vars_of_prog_ss) [] ``bir_vars_of_program ^prog``;
```

### DIFF
--- a/src/theory/bir-support/HolBASimps.sig
+++ b/src/theory/bir-support/HolBASimps.sig
@@ -3,9 +3,9 @@ sig
    include Abbrev
 
    (* Rewrite rules to determine statements of a program *)
-   val stmts_of_prog_ss : simpLib.ssfrag;
+   val STMTS_OF_PROG_ss : simpLib.ssfrag;
 
    (* Rewrite rules to determine variables of a program *)
-   val vars_of_prog_ss : simpLib.ssfrag;
+   val VARS_OF_PROG_ss : simpLib.ssfrag;
 
 end
--- a/src/theory/bir-support/HolBASimps.sig
+++ b/src/theory/bir-support/HolBASimps.sig
@@ -1,0 +1,11 @@
+signature HolBASimps =
+sig
+   include Abbrev
+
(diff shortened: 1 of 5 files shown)


andreaslindner:
I wouldn't add tactics there since it is not really clear what is needed in general and that's why I think an ssfrag is good there. You might also want to add the program definition theorem into the list of theorems of the SIMP_CONV or SIMP_TAC and then use a program constant in the term or goal you provide there.

Anyways, I added one test to the selftest and simplified the set further as opposed to the usage I suggested in the first comment for this pull request.

andreaslindner:
I don't see what you mean in this file, but I can change them to `STMTS_OF_PROG_ss` and `VARS_OF_PROG_ss` (like `WORD_ss` and `PRED_SET_ss`), right?

totorigolo:
> I don't see what you mean in this file

I searched in the HOL4 documentation if they mention some convention, and I found this (just the line I included in the link: 103).

> but I can change them to `STMTS_OF_PROG_ss` and `VARS_OF_PROG_ss` (like `WORD_ss` and `PRED_SET_ss`), right?

Yep :)

andreaslindner:
now i see, i didn't see the marked line. just blind today :)

totorigolo:
@andreaslindner What about adding `pred_setSimps.PRED_SET_ss` in the `ssfrags`? When I use them in their current state, I get duplicates, and adding this `ssfrag` removes them.

andreaslindner:
I'm thinking that this is not so good as this fragment is supposed to collect the set, not simplify the set. This simplification can also have negative performance implications. It is up to the use case whether you want to do everything at once or not - in the test I just did an EVAL afterwards to get the simplified set representation.

On top of this, I didn't find a way to combine fragments into new fragments. Maybe I was not looking for the right thing now that I think about it again.

PR #54, opened 2019-05-14
CI: Now greps for cheat and comment the PR

totorigolo:
What the title says.

If this PR is successful, we should see a comment from the new HolBA bot.

Closes: #52

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ dist: xenial
 env:
   global:
     - HOLBA_OPT_DIR=$HOME/cache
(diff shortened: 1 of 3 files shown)


totorigolo:
This PR adds ~40 sec + the time to boot a VM of latency to the CI process.

andreaslindner:
> This PR adds ~40 sec + the time to boot a VM of latency to the CI process.

I think there is a lot of variance in the job execution time that it is hard to tell this from one run. Also, I don't think it could add so much. Or how did you get the 40 seconds?

andreaslindner:
This is really nice, and the holba-bot icon also has something resembling horns ;)

andreaslindner:
i didn't check, could you explain the global secret property? i guess this is travis naming but it doesn't seem very secure if a secret is out in the public :)

totorigolo:
> This is really nice, and the holba-bot icon also has something resembling horns ;)

Oops, I just changed the profile picture :innocent: 


> i didn't check, could you explain the global secret property? i guess this is travis naming but it doesn't seem very secure if a secret is out in the public :)

This is the `HOLBA_BOT_GITHUB_TOKEN` environment variable, encrypted using the `travis` CLI.

> Looks good! could you mention the corresponding issue and fix the CONTRIBUTING.md file accordingly in an additional commit? or is there something missing for this?

I mentioned the issue in the PR (I edited), but I can also mention in the commit. Regarding the `CONTRIBUTING.md`, you're right I forgot it :)

PR #55, opened 2019-05-14
Obtaining the old Hoare triple from the new

### LINKED ISSUE #53: Duplicate theorems in bir_program_blocksTheory
I was casually strolling through `bir_program_blocksScript.sml` when I noticed `bir_get_current_block_block_pc_IS_SOME` and `bir_get_current_block_block_pc`. These appear to be exactly identical. Therefore, I suggest we get rid of one of them. `bir_get_current_block_block_pc_IS_SOME` is used 1 times throughout the entirety of HolBA and `bir_get_current_block_block_pc` is used a grand total of zero times. I recommend `bir_get_current_block_block_pc` for removal, but first I want to give any eventual friends of `bir_get_current_block_block_pc` a possibility to have a say in the matter.

didriklundberg:
I added new theorems and definitions to obtain an AssumptionViolated-free Hoare triple from the regular one, provided the program contains no Assume statements. Also, corrected typo in exec_to_labels HT.

The terminology is a little rough around the edges. I will likely make some clean-up commit that renames and moves some lemmata around, and makes the proofs nicer. Please provide feedback so that everything can be sorted out in one go.

I am also working on a proof procedure to obtain the result of an Assume-free BIR program (the definitions should be very straightforward to use, however).

EDIT: Added commit to resolve #53.

### DIFF
--- a/src/theory/tools/wp/bir_htScript.sml
+++ b/src/theory/tools/wp/bir_htScript.sml
@@ -23,11 +23,8 @@ open HolBACoreSimps;
 val _ = new_theory "bir_ht";
 
 val bir_stmtB_is_not_assume_def = Define `
-  (bir_stmtB_is_not_assume (BStmt_Declare v) = T) /\
-  (bir_stmtB_is_not_assume (BStmt_Assign v ex) = T) /\
-  (bir_stmtB_is_not_assume (BStmt_Assert ex) = T) /\
   (bir_stmtB_is_not_assume (BStmt_Assume ex) = F) /\
-  (bir_stmtB_is_not_assume (BStmt_Observe ec el obf) = T)
+  (bir_stmtB_is_not_assume _ = T)
 `;
 
 val bir_stmtB_not_assume_never_assumviol =
--- a/src/theory/tools/wp/bir_htScript.sml
+++ b/src/theory/tools/wp/bir_htScript.sml
@@ -1,21 +1,25 @@
 open HolKernel Parse boolLib bossLib;
 
-(* From /core: *)
+(* Theories from theory/bir: *)
 open bir_programTheory bir_typing_progTheory bir_envTheory
      bir_auxiliaryTheory bir_valuesTheory bir_expTheory
      bir_exp_immTheory bir_typing_expTheory bir_immTheory;
 
-(* From /theories: *)
+(* Theories from theory/bir-support: *)
 open bir_program_blocksTheory bir_program_terminationTheory
      bir_program_valid_stateTheory bir_exp_substitutionsTheory
      bir_bool_expTheory bir_program_env_orderTheory
      bir_program_multistep_propsTheory bir_exp_equivTheory
      bir_subprogramTheory;
 
+(* Local theories: *)
 open bir_wpTheory;
 
+(* Simplification sets from theory/bir: *)
 open HolBACoreSimps;
 
+(* This theory contains theorems for manipulation of Hoare
+ * triples not directly related to WP propagation. *)
 val _ = new_theory "bir_ht";
 
 val bir_stmtb_is_not_assume_def = Define `
@@ -26,8 +30,8 @@ val bir_stmtb_is_not_assume_def = Define `
   (bir_stmtb_is_not_assume (BStmt_Observe ec el obf) = T)
 `;
 
-val bir_stmtb_not_assume_never_assviol =
-  store_thm("bir_stmtb_not_assume_never_assviol",
+val bir_stmtb_not_assume_never_assumviol =
+  store_thm("bir_stmtb_not_assume_never_assumviol",
   ``!stmtb st obs st'.
     bir_stmtb_is_not_assume stmtb ==>
     (st.bst_status <> BST_AssumptionViolated) ==>
@@ -130,8 +134,8 @@ Cases_on `bir_dest_bool_val (bir_eval_exp b st.bst_environ)` >| [
 ]
 );
 
-val bir_stmtsB_not_assume_never_assviol =
-  store_thm("bir_stmtsB_not_assume_never_assviol",
+val bir_stmtsB_not_assume_never_assumviol =
+  store_thm("bir_stmtsB_not_assume_never_assumviol",
   ``!stmtsB l c st l' c' st'.
     bir_stmtsB_has_no_assumes stmtsB ==>
     (st.bst_status <> BST_AssumptionViolated) ==>
@@ -156,8 +160,8 @@ Cases_on `bir_state_is_terminated st` >| [
     FULL_SIMP_TAC std_ss [bir_exec_stmtB_exists]
   ) >>
   FULL_SIMP_TAC std_ss [LET_DEF] >>
-  IMP_RES_TAC bir_stmtb_not_assume_never_assviol >>
-  PAT_X_ASSUM ``!l'' c'' st''. blah``
+  IMP_RES_TAC bir_stmtb_not_assume_never_assumviol >>
+  PAT_X_ASSUM ``!l'' c'' st''. _``
               (fn thm => ASSUME_TAC
                 (SPECL [``(OPT_CONS obs_test l): 'a list``,
                         ``(SUC c):num``,
@@ -177,8 +181,8 @@ val bir_block_has_no_assumes_def = Define `
 `;
 
 (* TODO: Move somewhere else... *)
-val bir_exec_stmtE_not_assviol =
-  store_thm("bir_exec_stmtE_not_assviol",
+val bir_exec_stmtE_not_assumviol =
+  store_thm("bir_exec_stmtE_not_assumviol",
   ``!prog stmtE st.
       (st.bst_status <> BST_AssumptionViolated) ==>
       ((bir_exec_stmtE prog stmtE st).bst_status <>
@@ -272,8 +276,8 @@ ASSUME_TAC (SPECL [``h:'a bir_stmt_basic_t``, ``st:bir_state_t``]
 FULL_SIMP_TAC std_ss [LET_DEF]
 );
 
-val bir_block_not_assume_never_assviol =
-  store_thm("bir_block_not_assume_never_assviol",
+val bir_block_not_assume_never_assumviol =
+  store_thm("bir_block_not_assume_never_assumviol",
   ``!prog bl st l' c' st'.
     bir_block_has_no_assumes bl ==>
     (st.bst_status <> BST_AssumptionViolated) ==>
@@ -283,14 +287,14 @@ val bir_block_not_assume_never_assviol =
 FULL_SIMP_TAC std_ss [bir_block_has_no_assumes_def,
                       bir_exec_block_def] >>
 REPEAT STRIP_TAC >>
-IMP_RES_TAC bir_stmtsB_not_assume_never_assviol >>
+IMP_RES_TAC bir_stmtsB_not_assume_never_assumviol >>
 subgoal `?l_test c_test st'_test.
 	   bir_exec_stmtsB bl.bb_statements ([],0,st) =
              (l_test, c_test, st'_test)` >- (
   FULL_SIMP_TAC std_ss [bir_exec_stmtsB_exists]
 ) >>
 FULL_SIMP_TAC std_ss [LET_DEF] >>
-PAT_X_ASSUM ``!st' l' l c' c. blah``
+PAT_X_ASSUM ``!st' l' l c' c. _``
 	    (fn thm => ASSUME_TAC
 	      (SPECL [``st'_test:bir_state_t``,
 		      ``l_test: 'a list``,
@@ -319,7 +323,7 @@ Cases_on `bir_state_is_terminated st'_test` >| [
     ASSUME_TAC (SPECL [``prog:'a bir_program_t``,
                        ``bl.bb_last_statement:bir_stmt_end_t``,
                        ``st'_test:bir_state_t``]
-                      bir_exec_stmtE_not_assviol
+                      bir_exec_stmtE_not_assumviol
                ) >>
     Q.ABBREV_TAC `st' = bir_exec_stmtE prog bl.bb_last_statement
                                        st'_test` >>
@@ -353,10 +357,10 @@ FULL_SIMP_TAC std_ss [listTheory.INDEX_FIND_def] >>
 Cases_on `P h` >> (
   FULL_SIMP_TAC std_ss []
 ) >>
-PAT_X_ASSUM ``!i j P x. blah``
+PAT_X_ASSUM ``!i j P x. _``
             (fn thm => ASSUME_TAC (SPEC ``(SUC i):num`` thm)) >>
 REV_FULL_SIMP_TAC std_ss [] >>
-PAT_X_ASSUM ``!j' P' x'. blah`` (fn thm => IMP_RES_TAC thm) >>
+PAT_X_ASSUM ``!j' P' x'. _`` (fn thm => IMP_RES_TAC thm) >>
 REV_FULL_SIMP_TAC std_ss [arithmeticTheory.SUC_PRE]
 );
 
@@ -396,7 +400,7 @@ Induct_on `l` >| [
 		      )
 	       ) >>
     REV_FULL_SIMP_TAC std_ss [] >>
-    PAT_X_ASSUM ``!st' bl'. blah``
+    PAT_X_ASSUM ``!st' bl'. _``
 		(fn thm => ASSUME_TAC
 		  (SPECL [``st:bir_state_t``,
 			  ``bl:'a bir_block_t``] thm)
@@ -425,8 +429,8 @@ Cases_on `bir_state_is_terminated st'` >> (
 )
 );
 
-val bir_prog_not_assume_never_assviol =
-  store_thm("bir_prog_not_assume_never_assviol",
+val bir_prog_not_assume_never_assumviol_exec_block_n =
+  store_thm("bir_prog_not_assume_never_assumviol_exec_block_n",
   ``!prog st n bl ol nstep npc st'.
     (bir_get_current_block prog st.bst_pc = SOME bl) ==>
     bir_prog_has_no_assumes prog ==>
@@ -450,14 +454,14 @@ Induct_on `n` >| [
   rename1 `st'''.bst_status = BST_AssumptionViolated` >>
   rename1 `bir_exec_block prog bl st = (l',c',st')` >>
   rename1 `st''.bst_status = BST_AssumptionViolated` >>
-  PAT_X_ASSUM ``!prog st bl ol nstep npc st'. blah``
+  PAT_X_ASSUM ``!prog st bl ol nstep npc st'. _``
           (fn thm => ASSUME_TAC (SPECL [``prog: 'a bir_program_t``,
                                         ``st':bir_state_t``] thm)
           ) >>
   IMP_RES_TAC bir_prog_to_block_no_assumes >>
-  IMP_RES_TAC bir_block_not_assume_never_assviol >>
+  IMP_RES_TAC bir_block_not_assume_never_assumviol >>
   IMP_RES_TAC bir_exec_block_n_block >>
-  PAT_X_ASSUM ``!n. blah``
+  PAT_X_ASSUM ``!n. _``
           (fn thm => ASSUME_TAC
             (SPECL [``n:num``] thm)
           ) >>
@@ -472,7 +476,7 @@ Induct_on `n` >| [
       IMP_RES_TAC bir_exec_block_new_block_pc >>
       FULL_SIMP_TAC std_ss [optionTheory.IS_SOME_EXISTS] >>
       FULL_SIMP_TAC (std_ss++holBACore_ss) [] >>
-      PAT_X_ASSUM ``(ol,nstep,npc,st'') = blah``
+      PAT_X_ASSUM ``(ol,nstep,npc,st'') = _``
 	      (fn thm => ASSUME_TAC
 		(GSYM thm)
 	      ) >>
@@ -484,7 +488,7 @@ Induct_on `n` >| [
 
       FULL_SIMP_TAC (std_ss++holBACore_ss) [] >>
       IMP_RES_TAC bir_exec_block_n_REWR_TERMINATED >>
-      PAT_X_ASSUM ``!p n. blah``
+      PAT_X_ASSUM ``!p n. _``
 	      (fn thm => ASSUME_TAC
 		(SPECL [``prog:'a bir_program_t``, ``n:num``] thm)
 	      ) >>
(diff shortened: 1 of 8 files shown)


src/theory/tools/wp/bir_wpScript.sml
```diff
@@ -246,7 +246,7 @@ PAT_X_ASSUM ``bir_is_bool_exp_env s.bst_environ ex``
               (HO_MATCH_MP bir_bool_values thm)
             ) >>
 REV_FULL_SIMP_TAC std_ss [] >>
-FULL_SIMP_TAC std_ss [bir_not_val_true, bir_exec_stmtB_def,
+FULL_SIMP_TAC std_ss [bir_neg_val_true, bir_exec_stmtB_def,
```
totorigolo: Are you based on `master`? This [should be](https://github.com/kth-step/HolBA/blob/master/src/theory/bir-support/bir_exp_equivScript.sml#L99) `bir_not_equiv`.

totorigolo: https://travis-ci.com/kth-step/HolBA/jobs/200293490#L351

src/theory/tools/wp/bir_htScript.sml
```diff
@@ -0,0 +1,614 @@
+open HolKernel Parse boolLib bossLib;
+
+(* From /core: *)
```
totorigolo: This path changed, it should be `theory/bir` now. Same for `/theories` => `theory/bir-support.`.

Isn't `HolBACoreSimps` in `theory/bir`?

src/theory/tools/wp/bir_wpScript.sml
didriklundberg: Yep, this was a typo resulting from copy/pasting the work from an older branch.

src/theory/tools/wp/bir_htScript.sml
```diff
@@ -0,0 +1,614 @@
...
+     bir_program_valid_stateTheory bir_exp_substitutionsTheory
+     bir_bool_expTheory bir_program_env_orderTheory
+     bir_program_multistep_propsTheory bir_exp_equivTheory
+     bir_subprogramTheory;
+
+open bir_wpTheory;
+
+open HolBACoreSimps;
+
+val _ = new_theory "bir_ht";
```
totorigolo: Wouldn't `bir_hoare_triple` be easier to understand? Or you can at least add a comment in the top of the file to explain the acronym :)

src/theory/tools/wp/bir_htScript.sml
```diff
@@ -0,0 +1,614 @@
...
+
+val bir_stmtb_is_not_assume_def = Define `
+  (bir_stmtb_is_not_assume (BStmt_Declare v) = T) /\
+  (bir_stmtb_is_not_assume (BStmt_Assign v ex) = T) /\
+  (bir_stmtb_is_not_assume (BStmt_Assert ex) = T) /\
+  (bir_stmtb_is_not_assume (BStmt_Assume ex) = F) /\
+  (bir_stmtb_is_not_assume (BStmt_Observe ec el obf) = T)
+`;
+
+val bir_stmtb_not_assume_never_assviol =
```
totorigolo: This should be named `bir_stmtB_not_assume_never_assviol`. Also, as a French, I would prefer `assumviol` :sweat_smile:

src/theory/tools/wp/bir_htScript.sml
```diff
@@ -0,0 +1,614 @@
...
+);
+
+val bir_stmtsB_has_no_assumes_def = Define `
+  (bir_stmtsB_has_no_assumes [] = T) /\
+  (bir_stmtsB_has_no_assumes (h::t) =
+    (bir_stmtb_is_not_assume h) /\ (bir_stmtsB_has_no_assumes t)
+  )
+`;
+
+(* TODO: Move this elsewhere... *)
```
totorigolo: Can you explain why in the TODO? Same for the following ones.

src/theory/tools/wp/bir_htScript.sml
```diff
@@ -0,0 +1,614 @@
...
+                      bir_exec_block_def] >>
+REPEAT STRIP_TAC >>
+IMP_RES_TAC bir_stmtsB_not_assume_never_assviol >>
+subgoal `?l_test c_test st'_test.
+	   bir_exec_stmtsB bl.bb_statements ([],0,st) =
+             (l_test, c_test, st'_test)` >- (
+  FULL_SIMP_TAC std_ss [bir_exec_stmtsB_exists]
+) >>
+FULL_SIMP_TAC std_ss [LET_DEF] >>
+PAT_X_ASSUM ``!st' l' l c' c. blah``
```
totorigolo: IMO you should use `_` instead of `blah`. That's what I infer from https://hol-theorem-prover.org/kananaskis-12.release.html#incompatibilities , and I find it a bit cleaner.

src/theory/tools/wp/bir_htScript.sml
```diff
@@ -0,0 +1,614 @@
...
+    FULL_SIMP_TAC (std_ss++holBACore_ss) [],
+
+    FULL_SIMP_TAC std_ss [] >>
+    Cases_on `bir_env_update (bir_var_name b)
+                (bir_declare_initial_value (bir_var_type b))
+                (bir_var_type b)
+                st.bst_environ` >| [
+      FULL_SIMP_TAC std_ss [bir_state_set_failed_def] >>
+      RW_TAC std_ss [] >>
+      FULL_SIMP_TAC (std_ss++holBACore_ss) [],
```
totorigolo: You are using this simpset `(std_ss++holBACore_ss)` everywhere, shouldn't you create in once and use it instead of creating it for each use? This should also speed up the execution a bit.

src/theory/tools/wp/bir_htScript.sml
```diff
@@ -0,0 +1,614 @@
...
+
+val bir_prog_has_no_assumes_def = Define `
+  (bir_prog_has_no_assumes (BirProgram []) = T) /\
+  (bir_prog_has_no_assumes (BirProgram (h::t)) =
+    (bir_block_has_no_assumes h) /\
+    (bir_prog_has_no_assumes (BirProgram t))
+  )
+`;
+
+val test_lemma_3 = store_thm("test_lemma_3",
```
totorigolo: Should this be stored in the theory? (`store_thm` => `prove` ?)

src/theory/tools/wp/bir_htScript.sml
```diff
@@ -0,0 +1,614 @@
...
+      PAT_X_ASSUM ``!p n. blah``
+	      (fn thm => ASSUME_TAC
+		(SPECL [``prog:'a bir_program_t``, ``n:num``] thm)
+	      ) >>
+      FULL_SIMP_TAC (std_ss++holBACore_ss) []
+    ]
+  ) 
+]
+);
+(* Formulated in terms of bir_exec_to_labels: *)
```
totorigolo: If you need a comment to explain the name, maybe the naming is bad? :) Maybe you can include "exec" and "exec_to_labels" in the names?

src/theory/tools/wp/bir_htScript.sml
```diff
@@ -0,0 +1,614 @@
...
+REPEAT STRIP_TAC >>
+subgoal `bir_exec_to_labels_n ls prog st 1 =
+           BER_Ended ol c_st c_l st'` >- (
+  FULL_SIMP_TAC (std_ss++holBACore_ss) [bir_exec_to_labels_def]
+) >>
+IMP_RES_TAC bir_exec_to_labels_n_TO_bir_exec_block_n >>
+IMP_RES_TAC bir_prog_not_assume_never_assviol
+);
+
+val bir_exec_to_labels_no_av_triple_def = Define `
```
totorigolo: I find it hard to guess that `no_av` refers to the theorem in `bir_wpTheory`.

src/theory/tools/wp/bir_htScript.sml
andreaslindner: good idea, I think a comment suffices.

andreaslindner:
> I'm just commenting because I want to let @andreaslindner review, since it's his tool :)

Thanks for thinking about me, but we're working all together on this :)

src/theory/tools/wp/bir_htScript.sml
andreaslindner: hahaha, oh yes, i think we can avoid this wrong interpretation by renaming

src/theory/tools/wp/bir_htScript.sml
andreaslindner: yes, sound good

src/theory/tools/wp/bir_htScript.sml
andreaslindner: i'm not thinking about execution speed but it may be shorter and also become more readable. i will also try to apply this for future theories i think.

totorigolo: I think that the readability gain depends on the name, because in this case the code is small so it's useful to know that the proof is using `std_ss` with `holBACore_ss`, instead of (e.g.) `pure_ss`.

PS: shoudn't `holBACore_ss` be named `HOLBA_CORE_ss` (upper-case name)?

totorigolo:
> > I'm just commenting because I want to let @andreaslindner review, since it's his tool :)
> 
> Thanks for thinking about me, but we're working all together on this :)

It's also because I'm not really used to writing proofs, so I think that you review is way more valuable ;)

didriklundberg:
>     * Why did you create a new theory about Hoare triple, instead of adding these theorems in `bir_wpTheory`?

Because `bir_wpScript.sml` is already of humongous size. I like to maintain some degree of modularity and not keep all theorems in a single file. But of course, this subdivision can always be more precise. The intent of `bir_htTheory` is to contain HT manipulation theorems not directly related to the WP propagation mechanisms.

>     * Why did you add those `TODO`s instead of putting the theorems directly where they should be?

Not really for any strong reason, but I figured I could just as well receive comments first. I don't want to clutter existing theories with new theorems without making sure they go in the right place. Some of these theorems contain comments on where I intend to put them (so that we don't mess up commit history by moving them back and forth at PR review). So if @andreaslindner doesn't have any strong opinions I will just spread them out myself in a new commit.

src/theory/tools/wp/bir_htScript.sml
didriklundberg: > This path changed, it should be `theory/bir` now. Same for `/theories` => `theory/bir-support.`.

Check, will be addressed in a new commit.

> Isn't `HolBACoreSimps` in `theory/bir`?

Yes. All the theories, function libraries, syntax function libraries and simplification set libraries are also sorted by category, although this is hard to see from this example. I will write better comments so that this can be immediately seen.

By the way, I don't think it's necessary to always sort imports in this way, it's just tidy when you have 15+ of them.

src/theory/tools/wp/bir_htScript.sml
didriklundberg: "HT" is a very well established acronym for "Hoare triple", just like "WP" is for "weakest precondition", so I think this is OK. But I will add a comment describing the theory (IMO good practice anyway).

totorigolo: Oh, I didn't know about it. It's fine then :)

src/theory/tools/wp/bir_htScript.sml
didriklundberg: Okay, I will replace all instances of "assviol" with "assumviol", sparing posteriority from having to read ambiguous abbreviations.

src/theory/tools/wp/bir_htScript.sml
didriklundberg: The reason for this particular TODO is that the lemma is very general and as such does not belong in such a specialized theory as `bir_htTheory`. This theorem deals with fundamental properties of `bir_exec_stmtB` and IMO should be put in the same theory where this is defined (`bir_programTheory`). Similar for the following, but different theories.

The general thought has been detailed in a reply to your main comment.

totorigolo: I saw the reply to the main comment, no need to reexplain it here. This comment will be remarked as outdated if you do the spreading-out commit.

src/theory/tools/wp/bir_htScript.sml
didriklundberg: Yep, this is cleaner for sure, and it agrees with the wildcard operator in SML. I will change this.

Can you quote the text which suggests usage of _ over other dummy strings in the link?

totorigolo: > Can you quote the text which suggests usage of _ over other dummy strings in the link?

Sure! I wanted to link the following, even if it kinda doesn't apply here (emphasis mine):

> **The behaviour of the quoting versions of the tactics is also slightly different**: **they will always respect names** that occur both in the pattern and in the goal. Again, this is for consistency with similar functions such as `qspec_then`. This means, for example, that `qpat_assum 'a < b'` **will fail** if the actual theorem being matched is something `c < f a`. (This is because the pattern and the goal share the name `a`, so that the pattern is implicitly requiring the first argument to `<` to be exactly `a`, which is not the case.) This example would have previously worked if there was exactly one assumption with `<`. **The fix in cases like this is to use more underscores in one’s patterns**.

src/theory/tools/wp/bir_htScript.sml
didriklundberg: `(std_ss++holBACore_ss)` is super common in general. I suggest we abbreviate this to `std_holba_ss` directly in `HolBACoreSimps`.
 
> PS: shoudn't `holBACore_ss` be named `HOLBA_CORE_ss` (upper-case name)?

No idea, but the general pattern in HOL4 identifiers is that older stuff is in all caps, while newer stuff is all lowercase. HOL4 simpsets are all over the place.

What does @andreaslindner think?

totorigolo:
Btw, I focused on style and readability for a HolBA **user** in my review. Please ask if you have a specific question/concern :)

src/theory/tools/wp/bir_htScript.sml
didriklundberg: I have now renamed it to fit the terminology of `listTheory` where, IMO, this lemma should ideally be found. Yes, I am quite sure this should be stated with `store_thm`, but not in this theory.

@andreaslindner, what does the textbook say when we prove stuff that should really go into HOL4 mainline theories? As for this particular one, I suggest I drop it in `bir_auxiliaryScript.sml` (IMO this also applies to `FEVERY_FEVERY_DRESTRICT_thm` in `bir_wpTheory`, which I didn't author).

src/theory/tools/wp/bir_htScript.sml
totorigolo: > `(std_ss++holBACore_ss)` is super common in general. I suggest we abbreviate this to `std_holba_ss` directly in `HolBACoreSimps`.

Should be `std_holba_core_ss`, but I'm not sure if that's necessary. It's quite easy to create the ones we want in the top or our proofs. For example, in my current proof I included some `wordsTheory` and `pred_setTheory` ssfrags in my "`holba_ss`" simpset.


> > PS: shoudn't `holBACore_ss` be named `HOLBA_CORE_ss` (upper-case name)?

I was referring to [this](https://github.com/kth-step/HolBA/pull/50#pullrequestreview-237074280), I should have added the link. The idea is that by convention `ssfrag`s have upper-case names and `simpset` lower-case ones.

src/theory/tools/wp/bir_htScript.sml
didriklundberg: Okay, I am now renaming this and the previous theorem to reference the execution functions more explicitly.

src/theory/tools/wp/bir_htScript.sml
didriklundberg: Can you clarify?

src/theory/tools/wp/bir_htScript.sml
totorigolo: I was just saying this because of the name of the theorem.

src/theory/tools/wp/bir_htScript.sml
totorigolo: I said this because this definition is a copy-past from `bir_exec_to_labels_triple_def` in `bir_wpTheory` with the assumption violated disjunct removed. Maybe we should rename them:

- `bir_exec_to_labels_triple_def` => `bir_exec_to_labels_or_assumviol_triple_def`
- `bir_exec_to_labels_no_av_triple_def` => `bir_exec_to_labels_triple_def`

The motivation here is that users of the WP tool as a whole are more interested in `bir_exec_to_labels_no_av_triple_def` then in the current `bir_exec_to_labels_triple_def`.

src/theory/tools/wp/bir_htScript.sml
didriklundberg: > Should be `std_holba_core_ss`, but I'm not sure if that's necessary. It's quite easy to create the ones we want in the top or our proofs. For example, in my current proof I included some `wordsTheory` and `pred_setTheory` ssfrags in my "`holba_ss`" simpset.

True, but then we also have to contend with lots of identical abbreviations. The conclusion of this idea would be to redefine `std_holba_core_ss` in many files, since this combination is heavily used in a lot of places, now as well as going forward.

If we choose to go for a convenient shorthand, I also suggest we can even shorten it more, to `shc_ss` or something.

> I was referring to [this](https://github.com/kth-step/HolBA/pull/50#pullrequestreview-237074280), I should have added the link. The idea is that by convention `ssfrag`s have upper-case names and `simpset` lower-case ones.

Ah, I see. Thanks for the link.

Then I definitely agree we should rename `holBACore_ss`.

totorigolo: > True, but then we also have to contend with lots of identical abbreviations. The conclusion of this idea would be to redefine `std_holba_core_ss` in many files, since this combination is heavily used in a lot of places, now as well as going forward.

Ok I agree, if that's really that used then we should define it  in HolBACoreSimps as you said. Then I wonder if we should also create a simpset with pred_set and words inside, because when we are simplifying expressions we expect (at least I do) that `(1w: word1 <> 0w: word1)` is simplified to `T`.

src/theory/tools/wp/bir_htScript.sml
```diff
@@ -1,21 +1,25 @@
...
 open bir_wpTheory;
 
+(* Simplification sets from theory/bir: *)
 open HolBACoreSimps;
 
+(* This theory contains theorems for manipulation of Hoare
+ * triples not directly related to WP propagation. *)
 val _ = new_theory "bir_ht";
 
 val bir_stmtb_is_not_assume_def = Define `
```
totorigolo: You forgot to rename `stmtb` to `stmtB` in this one and the next theorem.

src/theory/tools/wp/bir_htScript.sml
didriklundberg: The reason for the current naming scheme is that `bir_exec_to_labels_triple` is mentioned much more often than `bir_exec_to_labels_no_av_triple`, which outside of its definition only exists in one theorem in HolBA. That said, I have no strong opinions on naming, so I can rename them.

src/theory/tools/wp/bir_htScript.sml
didriklundberg: It's been addressed now!

src/theory/tools/wp/bir_htScript.sml
didriklundberg: >Then I wonder if we should also create a simpset with pred_set and words inside, because when we are simplifying expressions we expect (at least I do) that `(1w: word1 <> 0w: word1)` is simplified to `T`.

If it's just isolated small stuff from wordsTheory, then maybe you can make use of the lemmata in `bir_auxiliaryTheory` (`word1_distinct` and `word1_neg`). But I feel this is a performance question, I have no idea how brute-force it is in practice to use the simpsets derived from wordsTheory over individual theorems.

I think this discussion is bigger than this PR, can you create an issue about future simpsets?

totorigolo: You're right, I created #57 and I'm resolving this.

src/theory/tools/wp/bir_htScript.sml
andreaslindner: Plus the ambiguity with assert :)

src/theory/tools/wp/bir_htScript.sml
andreaslindner: yes, maybe we should collect them in auxiliary and if you feel like it you mainline it and we could potentially remove it from auxiliary later at some point. I don't think we should develop more workflow for this than that :)

src/tools/wp/bir_htLib.sig
```diff
@@ -0,0 +1,10 @@
+signature bir_htLib =
+sig
+  include Abbrev
+
+  val bir_stmtb_is_not_assume_pp : term -> thm
```
totorigolo: You forgot the upper-case B again.

src/theory/tools/wp/bir_htScript.sml
andreaslindner: I don't totally agree to the point of interest of WP users. For compositions using assume statements for "function local" reasoning you need both. and this should be of interest for somebody who is proving a bigger program I think.

i can see arguments in both directions, for example:
- or assumviol is something more so it could be also more in the name
- the bir language itself has assume statements and therefore execution might end up in the assumviol state instead of running to the intended label

my suggestion is that we just keep it in mind for now and maybe defer the final naming to the point when wp becomes stable.

andreaslindner:
> The intent of bir_htTheory is to contain HT manipulation theorems not directly related to the WP propagation mechanisms.

I thought the same when I saw this pull request the first time. It makes totally sense to divide hoare triples from wp. One is the formalization of pre and post-conditions and the other is a mechanism making use of the formalizations.

src/theory/tools/wp/bir_htScript.sml
```diff
@@ -0,0 +1,513 @@
...
+open bir_wpTheory;
+
+(* Simplification sets from theory/bir: *)
+open HolBACoreSimps;
+
+(* This theory contains theorems for manipulation of Hoare
+ * triples not directly related to WP propagation. *)
+val _ = new_theory "bir_ht";
+
+val bir_stmtB_is_not_assume_def = Define `
```
andreaslindner: What about:
```
val bir_stmtB_is_not_assume_def = Define `
  (bir_stmtB_is_not_assume (BStmt_Assume ex) = F) /\
  (bir_stmtB_is_not_assume (_) = T)
`;
```

src/theory/tools/wp/bir_htScript.sml
andreaslindner: now i see that you already implemented it. good then :)

andreaslindner:
Oh, maybe a little test script for the new library could be useful.

src/theory/tools/wp/bir_htScript.sml
didriklundberg: Okay, I will change it.

src/tools/wp/bir_htLib.sig
didriklundberg: Addressed in new commit.

didriklundberg:
> Oh, maybe a little test script for the new library could be useful.

I'll see if I can get time to fix that later today.

PR #59, opened 2019-05-16
Fix bir_wp_comp_wps for BLE_Exp

didriklundberg:
An issue with `bir_wp_comp_wps` earlier was that it would just leave you with an exception if it ever encountered and tried to parse a BIR block with a `BLE_Exp` instead of a `BLE_Label`. These blocks represent function return statements. While this fix does not try to treat `BLE_Exp`s, it tells `bir_wp_comp_wps` to ignore them, which is OK since they are not needed anywhere in current usage.


didriklundberg:
This also needs to be put in dev_tutorial. @andreaslindner, can you see to that also after this PR is approved?

andreaslindner:
Yes, I will do the `dev_tutorial` business as soon as possible afterwards.

PR #60, opened 2019-05-18
Fixes and improvements for autoenv and Makefiles

andreaslindner:
This is a collection of fixes and improvements, the commit messages should help the review :)

### DIFF
--- a/scripts/setup/autoenv.sh
+++ b/scripts/setup/autoenv.sh
@@ -39,11 +39,15 @@ function is_holba {
 
   [[ -d "$1/scripts/setup" ]] \
       || { echo "bad scripts dir" > /dev/stderr; return $FALSE; }
+
+  [[ -f "$1/scripts/setup/autoenv.sh" ]] \
+      || { echo "no autoenv.sh" > /dev/stderr; return $FALSE; }
+
+  [[ "$(cat "$1/scripts/setup/autoenv.sh" 2> /dev/null | grep -c 'HOLBA')" -ge 1 ]] \
+      || { echo "bad autoenv.sh" > /dev/stderr; return $FALSE; }
+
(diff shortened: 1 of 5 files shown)


scripts/setup/autoenv.sh
```diff
@@ -39,11 +39,15 @@ function is_holba {
 
   [[ -d "$1/scripts/setup" ]] \
       || { echo "bad scripts dir" > /dev/stderr; return $FALSE; }
+
+  [[ -f "$1/scripts/setup/autoenv.sh" ]] \
+      || { echo "no autoenv.sh" > /dev/stderr; return $FALSE; }
+
+  [[ "$(cat "$1/scripts/setup/autoenv.sh" 2> /dev/null | grep -c 'HOLBA')" -ge 1 ]] \
```
totorigolo: Have you heard of [UUOC](http://catb.org/jargon/html/U/UUOC.html)? :stuck_out_tongue: But you can keep it.

Maybe you could do something like this instead:

```bash
HOLBA_AUTOENV_TOKEN="some-random-unique-token"
[[ $(grep -c "$HOLBA_AUTOENV_TOKEN" "$1/scripts/setup/autoenv.sh") == 1 ]] || fail
```

andreaslindner: No, I haven't heard of it - but now :). I will "double think" whether this could cause problems for the use cases we have and change it then.

totorigolo: The idea is just to avoid false positives, where a file exists and contains "HOLBA", but still isn't the right one. Maybe that's too paranoia, but then why are you greping for "HOLBA"? :stuck_out_tongue: And one plus of this method is that you could then update this token if you start having multiple versions of HolBA all around the place (the CI could `sed` it automatically for example).

andreaslindner: It's possibly a copy and paste mistake, I vaguely remember that I wanted to grep for autoenv.sh. But this is not much different anyways.

PR #62, opened 2019-05-21
Added infix in bir_wpLib

didriklundberg:
An issue with `bir_wpLib` is that the definitions for generated WPs can get overloaded. This is not problematic as long as you never generate and prove WPs for the same block labels in the same session. However, I think it would be nice to be able to do so.

This can be accomplished for example by adding an infix as an argument to `bir_wp_comp_wps` (and `bir_wp_comp_wps_iter_step2`) which ties these generated definitions to the particular WP.

In my estimation, this would also require changes to the following files (outside of the tutorial):

- `src/tools/wp/easy_noproof_wpLib.sml`
- `src/tools/wp/bir_wp_simpLib.sml`
- `src/tools/wp/examples/example-toy-pass.sml`
- `src/tools/wp/examples/test-toy.sml`
- `examples/aes/wp71-pass-test.sml`
- `examples/aes/wp-test.sml`
- `examples/aes/wp/aesM0WpsScript.sml`
- `examples/aes/wp/aesWpsScript.sml`
- `examples/aes/wp/aesM0SimpWpScript.sml`
- `examples/aes/wp/aesSimpWpScript.sml`

I am making this a draft PR to make it clear that we should discuss how to deal with this issue first.

### DIFF
--- a/src/tools/wp/bir_wpLib.sml
+++ b/src/tools/wp/bir_wpLib.sml
@@ -244,7 +244,7 @@ struct
 
   (* produce wps1 and reestablish bool_sound_thm for this one *)
   fun bir_wp_comp_wps_iter_step2 (wps, wps_bool_sound_thm) prog_l_thm
-				 ((program, post, ls), (label)) defs =
+				 ((program, post, ls), (label)) defs wp_infix =
     let
       val wps_id_suffix = label_to_wps_id_suffix label
 
@@ -280,7 +280,7 @@ struct
                                         GSYM bir_exp_and_def]
                                        wps1_thm; 
       val wps1 = (snd o dest_comb o snd o dest_eq o concl) wps1_thm
-      val new_wp_id = wps_id_prefix^wps_id_suffix
+      val new_wp_id = wps_id_prefix^wp_infix^wps_id_suffix
       val new_wp_id_var = mk_var (new_wp_id, bir_exp_t_ty)
       val new_wp_def =
         Define `^new_wp_id_var = ^(extract_new_wp wps1)`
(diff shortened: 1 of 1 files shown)


totorigolo:
I totally agree and :+1: this change. The current naming is `bir_wp_comp_wps_iter_step2_wp_[label]`. As you are suggesting, one way of doing it is by adding an infix. But we could also change the whole prefix. The current one is quite long and not very useful IMO.

What about:

```sml
fun bir_wp_comp_wps wp_prefix prog_def
                    ((wps, wps_bool_sound_thm), (wpsdom, blstodo))
                    (program, post, ls) defs =
  (* ... *)  
  val new_wp_id = wp_prefix ^ wps_id_suffix
```
Also, I would like having the prefix (of infix) parameter as the first one, as this facilitates reading the code.

```sml
val (wpsrec, wpsrec_bool_sound_thm) =
  bir_wp_comp_wps prog_thm ((wps, wps_bool_sound_thm),
                            (wpsdom, blstodo)
                            ) (program, post, ls) defs "wp71_pass_test";

(* versus *)

val (wpsrec, wpsrec_bool_sound_thm) =
  bir_wp_comp_wps "wp71_pass_test" prog_thm
                  ((wps, wps_bool_sound_thm), (wpsdom, blstodo))
                  (program, post, ls) defs;
```

</details>

andreaslindner:
Yes, I also agree of course. The faster somebody changes this the faster nobody can blame me for this embarrassment of mine :). Also the names are an unfortunate choice.

I am not sure whether we should go directly but we should rework this procedure to not make definitions. It's unneeded. One (in my eyes good) way to circumvent is the generation of triple theorems for the individual steps and storing them in a map. There is no need to do this in an HOL procedure with proven properties like we have it now. Actually a "quickerfix" would be to change the procedure to not output a whole HOL map (this causes rapid term growth and unnecessary garbage collector load) but rather just an option of the new mapping where we can reuse the existing proof to reflect only one change. This should be not too much work and could be a way to go.

andreaslindner:
I will look at the code later and also try to find you later.

andreaslindner:
note: this PR could become obsolete after PR #70

andreaslindner:
Now it is obsolete and I close it.

PR #70, opened 2019-05-29
[WIP] Merge tutorial changes to src

andreaslindner:
We need to check these changes.
I see two problems:
- we need to make sure that wpLib does not evaluate bir_exp_subst1 "away"
- we need to enable wpSimpLib to work without definitions, or make it work again when no definitions are used in the input expression

The first one and a partial/failed attempt on the second one are two commits in the branch `dev_tutorial_wpsubstrev`


totorigolo:
Just some thoughts:

* In "Removed definitions from generation of wp" (ccda083), why are you keeping commented-out code?
* The commit message "Fix error" (6c27bd4) could maybe be more explicit?
* "Adding an additional argument to WP initialization function" (0cf3761) fixes code, meaning that some code didn't compile before because of static errors. IMO this should be squashed to the breaking previous one.
* "Changes in bir_wpLib" => "bir_wpLib: clean up an ugly proof" or "bir_wpLib: remove useless REPEAT_MAX"?

But the most critical to me is that this PR is breaking the CI, are you fine with this going on `master`?

didriklundberg:
> But the most critical to me is that this PR is breaking the CI, are you fine with this going on `master`?

Yes, this must be fixed before. But I also think that @andreaslindner intends to do this, right?

Anyway, this seems to be missing things which would help in that regard, like for example commit 62f6996.

>     * "Adding an additional argument to WP initialization function" ([0cf3761](https://github.com/kth-step/HolBA/commit/0cf3761b7c9e21c34d3b7d227af039d0ab2922fc)) fixes code, meaning that some code didn't compile before because of static errors. IMO this should be squashed to the breaking previous one.

I'm not so sure, this depends on who authored the breaking commit. Also, the code that didn't compile was (likely) in the tutorial directory.

andreaslindner:
Yes, the intention was to fix it. I just opened it as a pull request directly with the todo list in the initial comment.

There is no specific commit breaking the CI, the problem is that the simplification is not implemented in a general way, at least not general enough that it can run on top of the definitionless wp procedure.

We could see this pull request as a rework task of the directory src/tools/wp.

didriklundberg:
> There is no specific commit breaking the CI

Well, commit 3795591 would break the CI without commit 62f6996 (which I think you accidentally left out from this PR).

> the problem is that the simplification is not implemented in a general way, at least not general enough that it can run on top of the definitionless wp procedure.

Yes, this is the main TODO item for this PR.

andreaslindner:
Maybe I should have just tagged this pull request as WIP or something. This is a pull request meaning: we should work on these and those commits to fix a problem until completion. I think this would avoided confusion - after all you're right, this is a "pull request". I used this because it has more value than an issue and I think we should do this fast.

Yes sure, there is always a breaking commit, i meant more the big picture or the incompleteness of the work for "removing definitions from this directory". Anyways, wpLib and wpSimpLib is interconnected and cannot be changed individually. Yes, if we change the interface of wpLib, and this causes tests to break - but this is not the main problem here.

I did not add all the commits because this first sequence here removes the definitions from wpLib (we wrote it together so this part is already somewhat checked) but does create the following problems:
- a runtime problem with wpSimpLib (which in turn assumes definitions to be there, and it's implementation also needs rework)
- evaluation of all bir_exp_subst1 functions
Both have to be addressed and only a part has been done in these additional two commits. I wanted to keep them separate so that they can be reviewed easier. But they are not solving the big problem. That is reworking wpSimpLib to be less of a first trial library.

totorigolo:
Btw, I think that you can just rename the failing tests `failing-test-xxx.sml` and move on, opening an issue to explain why it doesn't work and everything.

andreaslindner:
Yes, that's also an option. I would say we can do this if we really need and want to do something else first.

andreaslindner:
The last commit is essentially a realization of the suggestion Thomas made. We should fix the wp simplification separately and move on for now. With this I merge this pull request now.

PR #91, opened 2019-06-27
Use Python 3

### LINKED ISSUE #40: Change python2 to python3

totorigolo:
Chat the title says. Everything should still work fine (and the CI will confirm :slightly_smiling_face:).

Closes #40

### DIFF
--- a/doc/gen/dependencies.py
+++ b/doc/gen/dependencies.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python2
 
 
 import os
--- a/scripts/ci/post-comment-on-PR.sh
+++ b/scripts/ci/post-comment-on-PR.sh
@@ -29,7 +29,7 @@ declare COMMENT_BODY="$(cat)"
 
 # Escape the comment to respect JSON format
 json_escape () {
-    printf '%s' "$1" | python -c 'import json,sys; print(json.dumps(sys.stdin.read()))'
+    printf '%s' "$1" | python3 -c 'import json,sys; print(json.dumps(sys.stdin.read()))'
 }
 declare COMMENT_BODY=$(json_escape "$COMMENT_BODY")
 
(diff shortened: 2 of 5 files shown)


andreaslindner:
We need to check whether we can have `python3` also here:
```
scripts/ci/grep-cheat-comment-PR.sh:    printf '%s' "$1" | python -c 'import json,sys; print(json.dumps(sys.stdin.read()))'
scripts/setup/install_z3_src.sh:python scripts/mk_make.py --prefix=${Z3_DIR} --python --pypkgdir=${Z3_DIR}/bin/python
```

... and write python2 here (no need to fix this one...):
```
doc/gen/dependencies.py:#!/usr/bin/python
```

then this should close #40

totorigolo:
I think that you should have used `./script.py` instead :slightly_smiling:

https://github.com/kth-step/HolBA/blob/d2ca1032433cf96accabab9cfc85d7fe2f20ab71/scripts/setup/install_z3_src.sh#L55

andreaslindner:
okay, I could try this later. but what is the benefit of one over the other?

totorigolo:
Because then there is no need to say it twice, the shebang is enough :)

andreaslindner:
Do you mean creating a script.py file with this behavior? Otherwise I cannot find this file

totorigolo:
I just meant that you can just execute `./script.py` thanks to the [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)) :wink:

andreaslindner:
ah ok. but i'm still unsure whether i get you right. if you talk about the script scripts/mk_make.py. there is no shedingdong, see:
https://github.com/Z3Prover/z3/blob/master/scripts/mk_make.py

and even if there was one, it could be referring to just python, and i may want to be specific to use python3 for systems which are more legacy driven :)

totorigolo:
Right, I didn't notice that you don't own this file. Then, nevermind :innocent:

PR #92, opened 2019-06-27
examples/nic code

totorigolo:
I now, this PR is way too big. Anyway, it contains the `examples/nic` code that I did, and some related commit.

If this breaks because of changes in `master`, I'll be happy to maintain it a bit after I leave :slightly_smiling_face: I can also break this is smaller PR if needed.

Based on #94


totorigolo:
I will create a new PR that moves assum related stuff from the HT theory to the core, so this PR should be rebased on it when it gets merged (because of 1a3eeba).

PR #110, opened 2020-03-18
Add a new example for compositional binary hoare logic

andreaslindner:
Adds a simple example. Thanks for helping me, @didriklundberg.


didriklundberg:
No problem, @andreaslindner.

I also like that you clean up removed functions from `tutorial_compositionLib.sig`, but did you really add all the new ones for composition and manipulation of pre- and postcond?

PR #119, opened 2020-06-24
Introduce and id for observations

andreaslindner:
This adds assiciates a number to observations to enable multiple channels or similar.


didriklundberg:
Yes, I think this is very uncontroversial and straightforward.

The only thing is that I think we should write comments documenting the intent of all the fields of the Observe statement. First, the comments around line 435 in `bir_programScript.sml` must be modified according to the changes made. Secondly, I think new comments should be added around line 30 in `bir_programScript.sml` explaining the arguments of the Observe statement, possibly somewhere else an explanation of the semantics.

PR #124, opened 2020-07-17
SEFM cleanup

didriklundberg:
This removes the redundant `bir_triple` and simplifies the tutorial examples.


andreaslindner:
It seems the following test fails:
examples/tutorial/test-composition.sml

I guess that you changed something that changed some theorem/term shapes. Could you please inspect it in more detail and see how to fix it. Then, as usual, either adjust the expected test terms or fix the thing that causes wrong outputs.

didriklundberg:
> It seems the following test fails:
> examples/tutorial/test-composition.sml
> 
> I guess that you changed something that changed some theorem/term shapes. Could you please inspect it in more detail and see how to fix it. Then, as usual, either adjust the expected test terms or fix the thing that causes wrong outputs.

Fixed now!

PR #127, opened 2020-09-15
SEFM 2020 renaming

didriklundberg:
Files have been moved to agree with the link in the SEFM paper, also renamed.


andreaslindner (approved):
I like the renaming and find it more understandable this way.
Though, I didn't check all the details. After the tests passed we can simply merge it I think.

andreaslindner (approved):
Still approved :)

PR #129, opened 2020-09-21
Add RISC-V lifter support

didriklundberg:
This PR will add RISC-V support to the HolBA lifter.

I also changed the lifter self-tests a bit.


didriklundberg:
No idea why the commit dates have gone mad...

andreaslindner:
That's a nice set of additions. I took the liberty to add a commit to enable the new selftest files for the CI. I will add a few comments now, trying to use the Github interface to refer directly.

src/tools/lifter/bir_inst_liftingLib.sig
```diff
@@ -118,16 +118,20 @@ signature bir_inst_liftingLib = sig
   structure bmil_m0_BigEnd_Main : bir_inst_lifting;
 
 
-  (* M0_mod instance, little endian, process SP *)
+  (* M0 mod instance, little endian, process SP *)
```
andreaslindner, line 121: Just a comment, no need to do anything here. The underscore was intentional. It's not the model M0, but M0_mod.

src/theory/bir/bir_exp_memScript.sml
```diff
@@ -613,7 +613,12 @@ val bir_update_mmap_def = Define `
     (!mmap aty a.      (bir_update_mmap aty mmap a [] = mmap)) /\
     (!mmap aty a v vs. (bir_update_mmap aty mmap a (v::vs) =
                         bir_update_mmap aty (FUPDATE mmap ((bir_mem_addr aty a), v2n v)) (SUC a) vs))`;
-
+(* TODO: Order of updates in memory representation
+val bir_update_mmap_rev_def = Define `
```
andreaslindner, line 617: I'm curious, why is the rev update necessary?

andreaslindner:
I like your documentation in `src/theory/tools/lifter/bir_lifting_machinesScript.sml`. Nice!

andreaslindner:
I like what you did to the selftests! The changes are really good but it made me think and spell out thoughts that I had earlier. So, what comes is more a future thing I guess.

Probably we need even more tests to better understand which parts in the lifter break when we introduce changes that cause trouble. This goes a bit hand in hand with the mess that the file `src/tools/lifter/bir_inst_liftingLib.sml` presents. I think it is in a more prototype state because it lacks abstractions to make it digestible. It's essentially a horrible, huge, and monolithic chunk of code: a bunch of nested functions without understandable intermediate interfaces and proper testing. And neither are they easy to test, which is reflected in our additions, and specifically your additions now in terms of debug code comments. I suggest some additional abstraction level and intermediate test cases.

Similarly, we may want to think over the way how we plug in architecture models. Subdirectories are probably too much to ask because of Holmake, but maybe we can somehow separate them a bit better. And also continue the unification work you already put into the selftestLib for example to factor out common functionality to be reusable.

src/tools/lifter/bir_lifting_machinesLib_instances.sml
```diff
@@ -589,5 +609,327 @@ val _ = assert bmr_rec_sanity_check (m0_mod_bmr_rec_BigEnd_Main)
...
+ * until bir_nzcv_introsScript, riscv_stepScript,
+ * bir_riscv_extrasScript and bir_lifting_machinesScript can be
+ * modified to accommodate the RISC-V model. *)
+
+(* Type rewrites as a list of theorems (ARM8 also had rewrites
+ * for ``:ProcState``)... *)
+val riscv_REWRS = (
+  (type_rws ``:riscv_state``)
+);
+(* From M0:
```
andreaslindner, line 634: Why did you keep these copies in comments?

src/tools/lifter/bir_lifting_machinesLib_instances.sml
```diff
@@ -589,5 +609,327 @@ val _ = assert bmr_rec_sanity_check (m0_mod_bmr_rec_BigEnd_Main)
...
+    (SIMP_CONV std_ss
+               [bir_riscv_extrasTheory.word_add_to_sub_TYPES,
+                alignmentTheory.aligned_numeric,
+                wordsTheory.WORD_SUB_INTRO,
+                wordsTheory.WORD_MULT_CLAUSES]
+    )
+
+  (* TODO: bmr_extra_ARM8 is from bir_lifting_machinesTheory. *)
+(* From ARM8:
+  fun riscv_extra_THMS vn = let
```
andreaslindner, line 687: Here it is probably more clear what I meant in the last comment. If I don't see it wrong, then this comment starts with "From ARM8" and it is even the exact same like the function definition that follows after the comment. Why did you keep this?

didriklundberg:
> I like what you did to the selftests! The changes are really good but it made me think and spell out thoughts that I had earlier. So, what comes is more a future thing I guess.
> 
> Probably we need even more tests to better understand which parts in the lifter break when we introduce changes that cause trouble. This goes a bit hand in hand with the mess that the file `src/tools/lifter/bir_inst_liftingLib.sml` presents. I think it is in a more prototype state because it lacks abstractions to make it digestible. It's essentially a horrible, huge, and monolithic chunk of code: a bunch of nested functions without understandable intermediate interfaces and proper testing. And neither are they easy to test, which is reflected in our additions, and specifically your additions now in terms of debug code comments. I suggest some additional abstraction level and intermediate test cases.
> 
> Similarly, we may want to think over the way how we plug in architecture models. Subdirectories are probably too much to ask because of Holmake, but maybe we can somehow separate them a bit better. And also continue the unification work you already put into the selftestLib for example to factor out common functionality to be reusable.

I agree, we should maybe schedule a lifter architecture meeting to discuss this.

src/theory/bir/bir_exp_memScript.sml
didriklundberg: It's not necessary, it's older WIP material which was accidentally left in a comment :stuck_out_tongue:

I've removed it now, I'll push a new commit soon to address your comments.

src/tools/lifter/bir_inst_liftingLib.sig
didriklundberg: I've fixed it now.

src/tools/lifter/bir_lifting_machinesLib_instances.sml
didriklundberg: Yeah, more old TODO items and comments which had not been cleaned out. Fixed now.

didriklundberg (commented):
Okay, made some clean-up of comments. This should address your feedback.

PR #130, opened 2020-09-29
Move tutorial tools to main directory structure

didriklundberg:
This PR moves tools from the tutorial support directories to the other tools.

A few questions remain:

- What to do with `bir_lifter_simple_interfaceLib.sml` and `bir_lifter_interfaceLib.sml`? I suggest these be merged, while checking names of functions to be consistent.
- What to do with `bir_wp_interfaceLib.sml` and `easy_noproof_wpLib.sml` (currently not used)? Here, functionality seems to be more overlapping.
- SMT interface and symbolic execution stuff is left in the `support` and `support2` directories of the tutorial. We need to decide where to place these things.

We also need to change the main README to take these changes into account. I think that can be done when this is finalized.


andreaslindner (dismissed):
Nice restructuring collection. Could we make the adjustments to the README file in this PR and keep the three other bullet points for a new PR?

Otherwise, some comments:
- not for this PR, but general comment: we need more tests. at least some basic three tests per module. otherwise it's hard to notice if something breaks and also hard to find what exactly breaks from a change: e.g. backlifter and comp (at the moment they are indirectly tested by the tutorial test, but more fine grained testing would be helpful. even if tests are simple and really basic they help to spot some problems and narrow down the search for causes)
- backlifter sounds fancy :)
- maybe subprogram lib is not really a lifter thing. but maybe it is because it works together with programs that are lifted with the current version of the lifter. but then, when we change the lifter to produce multiple blocks per instruction, then this way of extracting subprograms doesn't work anymore. unsure. anyways, the reason why i comment on it is that the idea of having a subprogram lib doesn't sound like a "lifter thing" to me

didriklundberg:
Ok, just patched up `dependencies.py` a tiny bit but the output (both using `print_simple` flag and not) does not agree format-wise with the `dependencies.dot` originally added by @totorigolo in e572ec17f923ecb616ca74d3d2e599862a10bf02. @totorigolo, was this created manually or did you use some other tool I've missed?

totorigolo:
From the look of the file, I would say that I did create it manually. My rationale must have been that this directory structure does not change often, and it's super easy to fix the file.

didriklundberg:
> From the look of the file, I would say that I did create it manually. My rationale must have been that this directory structure does not change often, and it's super easy to fix the file.

Yes, that was what I suspected. Thanks for the quick reply.

andreaslindner:
Just to not keep you waiting: Should I give this one another review already to merge it or are you still working on it?

didriklundberg:
> Just to not keep you waiting: Should I give this one another review already to merge it or are you still working on it?

I pushed a commit with a few further changes to the README just now. You can review them, and maybe make an issue for fixing a new dependency graph if you want.

README.md
```diff
@@ -80,67 +80,76 @@ cd /path/to/{HOLBA_DIR}
...
-│     ├─ cfg: Control Flow Graph utilities
-│     ├─ exec: concrete execution
-│     ├─ lifter: proof-producing binary lifter
-│     ├─ pass: Passification utility
-│     ├─ wp: weakest precondition propagation
-│     └─ scamv: abstract side channel model validation framework
-└──── examples: to showcase HolBA
+├─ doc: Documentation material
+├─ examples: Showcasing HolBA
+│  ├─ aes: lifting + WP of AES on Cortex-M0
```
andreaslindner, line 92: I thought this is for ARM8 or both, or do I remember wrong?

README.md
```diff
@@ -80,67 +80,76 @@ cd /path/to/{HOLBA_DIR}
...
+│  ├─ bsl-wp-smt: Simplified BIR generation and SMT interface
```
andreaslindner, line 93: It does also have wp generation inside. In this way it tests the whole chain for simple example programs. Maybe we want to add one word somewhere to capture this fact?

README.md
didriklundberg: > I thought this is for ARM8 or both, or do I remember wrong?

Yes, it appears to be for both on a closer look, but only Cortex-M0 is mentioned in the filenames.

README.md
didriklundberg: > It does also have wp generation inside. In this way it tests the whole chain for simple example programs. Maybe we want to add one word somewhere to capture this fact?

Sure, you can suggest a change to this effect.

src/theory/models/l3mod/m0_mod_stepScript.sml
```diff
@@ -10,7 +10,7 @@ open combinTheory;
 val _ = new_theory "m0_mod_step";
 
 
-(* modified state with word representation of clock cyckle counter *)
+(* modified state with word representation of clock cycle counter *)
```
andreaslindner, line 13: Hahaha. Oops, nice find! 👍

andreaslindner (approved):
I think it looks good, and your idea to create an issue for the dependency graph is also good I think.

README.md
andreaslindner: Ah! Yes. Now I remember. That is probably because this one was added later while ARM8 was first the only one. And then I simply appended m0 to the new files.

README.md
andreaslindner: Not exactly one extra word, but what about:
Small BIR example programs to test Simplified BIR generation, WP propagation and SMT interface
or
Small BIR example programs to test BSL, WP, and the extended HolSMT
or append "libraries together" to the last one

andreaslindner (approved):
I just resolved the 3 merge conflicts stemming from the lifter wrapping. If the tests pass fine, then the last review should still apply.

andreaslindner:
Ahh. Probably because of the git file rename detection and similar path names, I merged wrong. I am fixing it now and will then check it properly.

andreaslindner (dismissed):
I checked and now everything is as it should be.

README.md
didriklundberg: I went with the first suggestion.

README.md
didriklundberg: Fixed now.

andreaslindner (approved):
Of course it's still very much approved 😃

PR #132, opened 2020-10-17
Merge latest scamv back to master

andreaslindner:
This requires cleanup work first, and also migration to HOL4 k13.


andreaslindner:
Done and double checked.

PR #134, opened 2020-10-24
Lifter selftest fixes

andreaslindner:
This renames the log files such that alphabetical sorting brings all the selftest logs together. It also fixes two bugs and the log outputs to make the selftest result meaningful again. Unfortunately, I overlooked this in the review and noticed it yesterday when looking at the unwrapped lifter libraries.


didriklundberg (approved):
No controversy here, I approve of the changes.

PR #135, opened 2020-10-24
Fix unwrapped lifter libraries

andreaslindner:
This is, unfortunately, a bigger issue that we postponed for too long. Some of the lifter libraries do not wrap their bindings, which makes everything available everywhere else. This missing encapsulation will soon cause problems with the bindings that this pull request addresses proactively and in a modular way one could say. Originally I thought the implications would be smaller, otherwise, I would have communicated and coordinated this change more before. But this is yet another reason why we need this containment of bindings rather sooner than later.

We should merge this into master rather quickly and merge it back into the other working branches to reduce merge conflicts and extra work because of this. Please review it as soon as possible such that another pair of eyes checked that really only bindings are changed, except for some irrelevant small adjustments.


didriklundberg (approved):
OK since you fixed the conflicts, but please do make an issue to clean up all of these added dependencies so that we don't just leave them behind everywhere.

andreaslindner:
> OK since you fixed the conflicts, but please do make an issue to clean up all of these added dependencies so that we don't just leave them behind everywhere.

Absolutely, I thought the same when I was doing this. First, I started with cleaning up every single one of them (except for the lifter helper, because I thought that you are refactoring these parts probably anyways), and after a while, I got very tired of it :).

Issue #103, opened 2019-10-17
Generation of documentation


palmskog:
Many kinds of documentation of a project like this are useful. Here is an attempt to break down different kinds of documentation and provide hints about how they can be generated:

- Project metadata, including short descriptions of what the project is about (1 sentence, 1 paragraph, ...)
see [example](https://github.com/coq-community/aac-tactics) from the Coq community
- Changelog summarizing how the project has evolved, for example using the [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) format
 see [example](https://github.com/coq-community/chapar/blob/master/CHANGELOG.md) from the Coq community
- Learning resources, such as listing of examples and step-by-step tutorials
 see [example](https://github.com/coq/coq/wiki#coq-books-and-tutorials) from the Coq community
- Website which organizes and contains the all of the above, preferably mostly auto-generated
 see [example](https://math-comp.github.io) from the Coq community

This issue is connected to #102, since releases usually deserve special documentation.

PR #142, opened 2020-11-09
New build passing badge

didriklundberg:
I added a new CI badge to the top. This looks nice, I think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-[![Build Status](https://travis-ci.com/kth-step/HolBA.svg?branch=master)](https://travis-ci.com/kth-step/HolBA)
-
 # HolBA - Binary analysis in HOL 
+
+[![Build Status](https://travis-ci.com/kth-step/HolBA.svg?branch=master)](https://travis-ci.com/kth-step/HolBA)
 
 Be sure to check out the Wiki, which contains some useful general information about HolBA.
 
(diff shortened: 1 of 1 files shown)


didriklundberg:
@andreaslindner , could you take a look at the reason for the below when you have time: https://github.com/kth-step/HolBA/blob/846fb86d36294af3daeecda87c88d741592a37b8/scripts/setup/env_config_gen.sh#L42-L43

Ironically, this commit caused the build to fail due to this check.

andreaslindner:
I didn't add the badge because it is almost non-sense. Browsers cache these images and thus a failing check (because of the weird deadlocks that sometimes happen for example) may stay preserved for very long.

As I see it this is a non-sense feature that everybody adds. Especially since it is added by more maintained projects that keep it always passing. The information content of this is near zero and because many add it, it has kind of a status value.

That being said, I am not against if you want to have it.

You can simply change the check in `HolBA/scripts/setup/env_config_gen.sh` in such a way that it still validates in a meaningful way that it indeed found the root directory of HolBA but works with the new changes. It doesn't need to check that the first line of the README starts with `# HolBA`. This was just a nice extra factor adding to the previous ones in this file.

didriklundberg:
I changed the position of the badge, which made the "bad README" check pass. I've seen both positions in other repositories, I think this also looks nice.

I was going through some earlier issues earlier and did this very quick PR in response to #103, I didn't recall any opinions on this matter from previously.

Personally, I think the README should look nice and be useful for visitors, so even if things don't make a lot of sense for us (we know builds on main branch pass CI) it might still serve a purpose in the README file. There are also some other badges we might consider adding later, like for example the Zulip chat link if we should choose to migrate to that platform.

andreaslindner:
Yes, that could be also useful if we plan to have a channel for the general public or similar.

andreaslindner (approved):
Looks good.

PR #143, opened 2021-02-04
Move to Kananaskis-14

didriklundberg:
The changes contained required to keep HolBA up-to-date with Kananaskis-14 are very small.

`make main` and `make tests` currently compile without errors, we will have to discuss if other things should be ticked off.

### DIFF
--- a/src/tools/lifter/selftest_arm.sml
+++ b/src/tools/lifter/selftest_arm.sml
@@ -5,6 +5,7 @@ open bir_inst_liftingLibTypes;
 open bir_inst_liftingHelpersLib;
 open PPBackEnd;
 
+open selftest_styleLib;
 open selftestLib;
 
 (******************)
--- a/src/tools/lifter/selftest_arm.sml
+++ b/src/tools/lifter/selftest_arm.sml
@@ -926,11 +926,7 @@ in () end;
 val arm8_expected_failed_hexcodes:string list =
 [
    "9BC37C41" (* umulh x1, x2, x3 lifting of ``Imm64 ((127 >< 64) (w2w (ms.REG 3w) * w2w (ms.REG 2w)))`` failed *),
-   "9B437C41" (* smulh x1, x2, x3 lifting of ``Imm64 ((127 >< 64) (sw2sw (ms.REG 3w) * sw2sw (ms.REG 2w)))`` failed *),
-   "DAC01441" (* clz x1, x2 lifting of ``Imm64 (n2w (CountLeadingZeroBits (ms.REG 2w)))`` failed *),
-   "5AC01441" (* clz w1, w2 lifting of ``Imm64 (n2w (BITS 31 0 (CountLeadingZeroBits (w2w (ms.REG 2w)))))`` failed *),
-   "DAC01041" (* cls x1, x2 lifting of ``Imm64 (n2w (CountLeadingSignBits (ms.REG 2w)))`` failed *),
-   "5AC01041" (* cls w1, w2 lifting of ``Imm64 (n2w (BITS 31 0 (CountLeadingSignBits (w2w (ms.REG 2w)))))`` failed *)
+   "9B437C41" (* smulh x1, x2, x3 lifting of ``Imm64 ((127 >< 64) (sw2sw (ms.REG 3w) * sw2sw (ms.REG 2w)))`` failed *)
 ];
 
 val _ = if (not test_arm8) then () else let
--- a/src/tools/lifter/selftest_arm8.log
+++ b/src/tools/lifter/selftest_arm8.log
@@ -3,37 +3,37 @@ MANUAL TESTS - ARMv8
 
 8B020020 (add x0, x1, x2) @ 0x10030 - OK
  []
-|- bir_is_lifted_inst_prog arm8_bmr (Imm64 65584w) (WI_end 0w 16777216w)
-     (65584w,[32w; 0w; 2w; 139w])
+|- bir_is_lifted_inst_prog arm8_bmr (Imm64 0x10030w) (WI_end 0w 0x1000000w)
+     (0x10030w,[32w; 0w; 2w; 139w])
      (BirProgram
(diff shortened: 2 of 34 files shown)


src/tools/lifter/selftest_arm.sml
```diff
@@ -925,11 +926,7 @@ in () end;
 val arm8_expected_failed_hexcodes:string list =
 [
    "9BC37C41" (* umulh x1, x2, x3 lifting of ``Imm64 ((127 >< 64) (w2w (ms.REG 3w) * w2w (ms.REG 2w)))`` failed *),
-   "9B437C41" (* smulh x1, x2, x3 lifting of ``Imm64 ((127 >< 64) (sw2sw (ms.REG 3w) * sw2sw (ms.REG 2w)))`` failed *),
```
andreaslindner, line 928: I don't understand why this happened? Does something work that didn't work before?

src/tools/lifter/selftest_arm8.log
```diff
@@ -3,37 +3,37 @@ MANUAL TESTS - ARMv8
 
 8B020020 (add x0, x1, x2) @ 0x10030 - OK
  []
-|- bir_is_lifted_inst_prog arm8_bmr (Imm64 65584w) (WI_end 0w 16777216w)
```
andreaslindner, line 6: Oh boy, did this change two times now? from k12 to k13 and then back from k13 to k14? Haha. At least I have this déjà vu feeling.

didriklundberg: Yes, I think you are correct.

src/tools/scamv/proggen/bir_prog_gen_randLib.sml
```diff
@@ -149,7 +149,7 @@ struct
      alphaList @ numList @ (regExLib.literalList "'_") @ specialChar @ [LITERAL #" ", LITERAL #"\t", LITERAL #"\n"]
  val alphabet_r = ALTERNATION (identifierList);
 
- val Any = STAR alphabet_r;
```
andreaslindner, line 152: Just a question. Why was this change necessary?

src/tools/lifter/selftest_arm.sml
didriklundberg: Four of the hexcodes that were listed as "expected to fail" actually worked. I don't think this was fixed by k14, since they were previously listed as "fixed" in the log. I removed them from the list of hexcodes expected to fail, which explains the changes.

src/tools/scamv/proggen/bir_prog_gen_randLib.sml
didriklundberg: There was some kind of clash with something else called "Any". Ugly, I know, please suggest a better fix if you have anything in mind.

andreaslindner:
I was thinking that we probably want to also add the latest stable polyml and make it the standard build. and drop the older one we have in the CI. After testing this locally, we need to change the CI cache key so that the CI rebuilds the cache with these combinations. The cache key update we need to do either way, but we could combine this now nicely I think.

src/tools/lifter/selftest_arm.sml
andreaslindner: great. nice!

src/tools/scamv/proggen/bir_prog_gen_randLib.sml
andreaslindner: No, I think we should leave it at that. I was just curious how this can clash because I thought it would need to be a keyword rather than an identifier for this to happen.

didriklundberg:
> I was thinking that we probably want to also add the latest stable polyml and make it the standard build. and drop the older one we have in the CI. After testing this locally, we need to change the CI cache key so that the CI rebuilds the cache with these combinations. The cache key update we need to do either way, but we could combine this now nicely I think.

Sure, good idea. Will you fix that and push to this branch or should I do it?

andreaslindner:
This is the latest polyml:
https://github.com/polyml/polyml/releases/tag/v5.8.1

And we could simply drop 5.7 now.

And this is where we need to change the CI cache key:
https://github.com/kth-step/HolBA/blob/042cb4cf0ea7d05b0769f950f444c5c2e1159649/.github/workflows/build.yaml#L35

Just add the hol4 version to it after polyml is adapted. The line could then be something like:
key: os-${{ runner.os }}_polyml-${{ matrix.polyml }}_hol4-k14

andreaslindner:
> > I was thinking that we probably want to also add the latest stable polyml and make it the standard build. and drop the older one we have in the CI. After testing this locally, we need to change the CI cache key so that the CI rebuilds the cache with these combinations. The cache key update we need to do either way, but we could combine this now nicely I think.
> 
> Sure, good idea. Will you fix that and push to this branch or should I do it?

Both works for me. If you like to do it, then go ahead. If you like me to do it, I can do that also.

didriklundberg:
> > > I was thinking that we probably want to also add the latest stable polyml and make it the standard build. and drop the older one we have in the CI. After testing this locally, we need to change the CI cache key so that the CI rebuilds the cache with these combinations. The cache key update we need to do either way, but we could combine this now nicely I think.
> > 
> > 
> > Sure, good idea. Will you fix that and push to this branch or should I do it?
> 
> Both works for me. If you like to do it, then go ahead. If you like me to do it, I can do that also.

I suggest you do it, you know more about the CI.

andreaslindner:
Wait, is this info still up to date?
https://github.com/kth-step/HolBA/blame/master/README.md#L64

didriklundberg:
> Wait, is this info still up to date?
> https://github.com/kth-step/HolBA/blame/master/README.md#L64

Hmm... I think [5.7.1 is still the latest version packaged for Ubuntu](https://packages.ubuntu.com/focal/polyml). So if we upgrade to 5.8.1 we need to keep that in mind when changing this line.

andreaslindner (approved):

PR #145, opened 2021-10-07
Fix printf warnings in run-test.sh

didriklundberg:
Just a quick fix for the pesky `printf` warning you get when running tests.


src/tools/lifter/selftestLib.sml
```diff
@@ -24,33 +28,16 @@ end;
 *)
   local
 
-(* these dependencies probably need cleanup *)
```
andreaslindner, line 27: Nice cleanup :)

scripts/setup/install_hol4_latest.sh
```diff
@@ -15,15 +15,13 @@ source "${SETUP_DIR}/env_config_gen.sh" "${OPT_DIR_PARAM}"
 
 ##################################################################
 
-
-# make polyml binaries and libraries available
+# use a default polyml version if it is not specified in the environment
 POLY_VERSION="v5.8.1"
```
andreaslindner, line 19: should we change the default here to 5.9 as well?

scripts/setup/install_hol4.sh
```diff
@@ -15,22 +15,25 @@ source "${SETUP_DIR}/env_config_gen.sh" "${OPT_DIR_PARAM}"
 
 ##################################################################
 
-
-# make polyml binaries and libraries available
+# use a default polyml version if it is not specified in the environment
 POLY_VERSION="v5.8.1"
```
andreaslindner, line 19: Also here, should we change the default here to 5.9?

.github/workflows/build.yaml
```diff
@@ -17,10 +17,15 @@ jobs:
 
     strategy:
       matrix:
-        polyml: ['v5.7.1', 'v5.8.1']
+        polyml: ['v5.7.1', 'v5.9']
```
andreaslindner, line 20: I thought you wanted to remove 5.7.1 and keep 5.8.1 as an alternative to the latest stable version?

andreaslindner (approved):
Otherwise I didn't notice anything, looks good!

scripts/setup/install_hol4_latest.sh
didriklundberg: Ah yes, I missed that. I'll add it.

scripts/setup/install_hol4.sh
didriklundberg: Yes.

.github/workflows/build.yaml
didriklundberg: No, not currently at least. I think it makes more sense to have 5.9 (which is the latest release), and 5.7.1 (which is the latest packaged for LTS Ubuntu). The reason for 5.9 is obvious, I think 5.7.1 makes the most sense as an alternative since this represents people using their default Poly/ML installation. I could be wrong though.

andreaslindner: Oh yes, sounds very reasonable.

andreaslindner (approved):
Still good :)

PR #146, opened 2021-10-08
Adding indirect jump resolution

didriklundberg:
This PR should put @adrwes work on indirect jump resolution into the main HolBA repository.

@adrwes, anything we should know before reviewing this?

### DIFF
--- a/examples/ijr/Makefile
+++ b/examples/ijr/Makefile
@@ -1,0 +1,18 @@
+# For compilation and disassembling
+HOLBA_DIR = ../..
+include $(HOLBA_DIR)/scripts/setup/env_check_gcc_arm8.mk
+CROSS =$(HOLBA_GCC_ARM8_CROSS)
+CFLAGS_BAREMETAL = -Iinc -ggdb3 -std=gnu99 -Wall -fno-builtin # -O1
+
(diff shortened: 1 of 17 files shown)


adrwes:
The work mainly consists of two proof-producing procedures: 
* `resolve_indirect_jumps` which "removes" indirect jumps in the input program based on a **proposal of indirect jump targets**. It basically just evaluates `resolve_fully_n`. `resolve_indirect_jumps` is not very efficient since  `resolve_fully_n` transforms the input program in multiple passes.
* `transfer_contract` which transfers a contract from the transformed program to the original one. It checks some conditions and instantiates the main correctness theorem of `resolve_fully_n` (`resolve_fully_n_contract_transfer`). (`prove_and_transfer_contract` also tries to prove the contract using existing infrastructure and logs some things)

The file `examplesScript.sml` contains examples of how the proof-producing procedures can be used.

andreaslindner (approved):
I'm not sure what to comment here. Most of the changes are new files in a new examples directory. We could debate the use of new HOL4 syntax and the use of unicode characters but I feel like this doesn't matter much, especially since it's in this new example directory.

So, looks good to me.

PR #151, opened 2022-05-09
a new eval function with mem rw

ningdongywq:
A new function bir_eval_mem_exe


didriklundberg (approved):
Everything looks good, and it compiles.

PR #152, opened 2022-05-31
Partial correctness unstructured Hoare logic

didriklundberg:


andreaslindner (approved):
Of course this mixes styles in the codebase, because you use the "new" HOL4 syntax. But I think this is no big deal. And otherwise I see nothing controversial from a high-level point of view.

PR #161, opened 2024-02-13
Parameterize DA lifter and add RISC-V tutorial

palmskog:
This is a revival and reimagining of https://github.com/kth-step/HolBA/compare/dev_riscv_tutorial which adds two new simple RISC-V examples. This should allow people to actually try out HolBA on RISC-V code.


examples/riscv/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+```shell
+$ git clone https://github.com/riscv/riscv-gnu-toolchain
+```
+
+Install the prerequisites, e.g., on Ubuntu:
+
+```shell
+$ sudo apt-get install autoconf automake autotools-dev curl python3 python3-pip libmpc-dev libmpfr-dev libgmp-dev gawk build-essential bison flex texinfo gperf libtool patchutils bc zlib1g-dev libexpat-dev ninja-build git cmake libglib2.0-dev
+```
```
didriklundberg, line 15: I understand usage of `$` to distinguish between the system shell and other command prompt interfaces, however here it is used only for the two first commands and nowhere else. In this particular README, there's also no other prompt to confuse the system shell with. I suggest to remove them, unless a good reason says otherwise.

didriklundberg (changes requested):

examples/riscv/README.md
palmskog: Done in 26617a7

didriklundberg (approved):

PR #165, opened 2024-03-04
Additional opens in bir_m0_extrasScript

didriklundberg:
Fixes the issues with Cortex-M0-lifting. The problem was that `R_name` from `m0_stepTheory` was treated as a free variable when the theory in question wasn't opened. Since this happened in a forward proof, it wasn't discovered during regular compilation.


palmskog (approved):
Can confirm this makes lifting work again for m0 in `src/tools/lifter/selftest_arm.sml`.

PR #175, opened 2024-03-18
Removed Z3 dependency in main theories

didriklundberg:


palmskog (approved):
Tested this locally, works fine.

PR #176, opened 2024-04-23
Update riscv incr example: end-to-end proof

andreaslindner:


examples/riscv/test-riscv.sml
```diff
@@ -27,15 +28,34 @@ val _ = print_and_check_thm
...
      riscv_swap_pre riscv_swap_post``;
+  
+val _ = print_and_check_thm
+  "incr RISC-V lift theorem"
+  bir_incr_riscv_lift_THM
+  ``
+  bir_is_lifted_prog riscv_bmr (WI_end (0w : word64) (8w : word64))
+   bir_incr_progbin
+   (bir_incr_prog : 'observation_type bir_program_t)
+  ``;
```
palmskog, line 61: How about adding a test for the final `incr` theorem? Some intermediate theorem could also make sense?

palmskog (commented):
Some quick comments, I assume stuff like indenting will be done during cleanup.

palmskog:
@andreaslindner so can I do a small pass on this PR and then approve and merge it? Would be very useful to have some end-to-end proofs as basis, not least when we have a port to the new HOL4.

palmskog (approved):

PR #178, opened 2024-04-29
Refactoring of symbolic execution automation and application to RISC-V examples

palmskog:
This PR defines a simple signature for running symbolic execution on BIR programs and obtaining a resulting general theorem (which holds soundness of the execution, etc.).

The signature function is applied to three examples obtained from RISC-V programs: `incr`, `mod2` and `swap`. There is also cleanup of the symbolic execution theory of each example.


examples/riscv/common/bir_symbLib.sml
```diff
@@ -0,0 +1,53 @@
...
+  open bitTheory;
+  open birs_stepLib;
+  open birs_composeLib;
+  open birs_driveLib;
+  open birs_auxTheory;
+in
+
+fun bir_symb_analysis bprog_tm
+ birs_state_init_lbl birs_stop_lbls
+ bprog_envtyl birs_pcond =
```
andreaslindner, line 17: maybe we could remove the bprog_envtyl and simply compute it from the program "internally"?

andreaslindner: or do we need it outside currently?

andreaslindner:
as far as i can tell it looks good

andreaslindner (approved):
Looks good

PR #182, opened 2024-06-11
always use HolBA_HolSmtLib.Z3_ORACLE_TAC instead of bundled Z3_ORACLE_TAC

palmskog:
This could be the reason for Z3 issues @didriklundberg, or what do you think?


didriklundberg (approved):
Regardless of whether this fixes the problem in question, it seems like sensible clean-up.

PR #187, opened 2024-08-16
CV Computation examples

jules-timmerman:
This is the work I have done regarding cv computation on a fragment of BIR, as well as multiple equivalent definitions of the semantics.
Tell me if you have any idea on more documentation I should add or things should change in the code.
I might add a few things later so I’ll keep it as a draft for now


palmskog:
@jules-timmerman quick small comments:
- no need for a separe `.gitignore`, the global one in HolBA covers everything
- there shouldn't be a separate `.holpath` file, you may have to address relativeness to `HOLBADIR` instead, which is in repo root

examples/compute/Holmakefile
```diff
@@ -0,0 +1,11 @@
+CLINE_OPTIONS = -r
+
+INCLUDES = $(HOLBADIR)/examples/compute/src/theory \
+					 $(HOLBADIR)/examples/compute/test \
+					 $(HOLBADIR)/examples/compute/examples \
```
palmskog, line 5: Could we make running the tests a task in this `Holmakefile`? See https://github.com/kth-step/HolBA/blob/master/examples/riscv/Holmakefile

examples/compute/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Remarks
+
+Here are some various remarks about the project :
+  - The `eval` semantics of program isn’t clean.
+  - The conversions in `shared/bir_cv*Lib.sml` should create a theorem by using the input term. Here, we are creating the output theorem from the right side and rewriting it to make the input term appear (but there is no guarantee).
+  - Implementation in the actual HolBA project can be done in two ways :
+    - You either keep the project structure the same and add support for unimplemented operations in the `bir_cv` representation.
+    The representation of HolBA and the `compute` one here are really similar (to a few differences like enforcing typing in environment manipulation).
+    - The other way would be to change the BIR representation in HolBA and use the `bir_cv` representation instead. This would remove translation cost but a lot more rewriting would be necessary.
```
palmskog, line 60: Can we move the TODO items here as well? Maybe under the label "Future work".

palmskog (commented):
Pure tests should ideally live outside of HOL4 theories (`new_theory`/`export_theory`). You can use `Theorem[local]` or the `prove` function to build theorems that are not stored in the HOL4 database. Ideally tests can be rerun again and against using `Holmake test`.

examples/compute/Holmakefile
jules-timmerman: Sorry I had only made it available inside the `examples/compute/examples/`. https://github.com/kth-step/HolBA/pull/187/commits/83c77721111cca5ef28f79a2bc34bddfd340f6b5 should fix this

examples/compute/README.md
jules-timmerman: Done in https://github.com/kth-step/HolBA/pull/187/commits/08cad5c722c930c448acfde3fb3bb691ece67527

examples/compute/Holmakefile
jules-timmerman: Hmmm I’m sorry I misunderstood (again…). You meant the tests should be run from the `Holmakefile`, not the examples (Could be both obviously). 
I’ll check what I can do regarding tests

palmskog: @jules-timmerman building the `examples/*` stuff is already done via the `-r` CLINE_OPTION. It's fine that the examples are in their own `*Script.sml` files and get built like HOL4 theories.

However, I argue that small tests should not be in `*Script.sml` files, but in regular files called something like `test-stuff.sml`, `test-more-stuff.sml`, that live in a relevant directory, which could here be `$(HOLBADIR)/examples/compute`. This ensures they [the tests] are not persisted as theories and can be run from scratch again and again via `Holmake test`.

palmskog: Recall that as long as you open HOL4 core modules:
```sml
open HolKernel boolLib Parse bossLib;
```
you can define theorems and the like using the `prove` function` or `Theorem my_local_thm[local] ...`, without being inside a theory.

If you need some `Definition` to define the local theorem, I argue this definition should live in some non-test HOL4 theory and `open`ed in the `test-*.sml` file.

Does that make it clearer?

jules-timmerman: Okay I understand now thank you very much for your explanation. I’ll try changing that. Sorry for the ugly git history

palmskog: Please feel free to make as many commits as you want, I think we will squash PR to make it a single commit in the HolBA history. The private development repo can have the development history if we need it.

jules-timmerman: Okay I moved the tests in a single file in https://github.com/kth-step/HolBA/pull/187/commits/5c858d2c1019548f6c74fd74ff33ccd9b82fca6e. I think this should be what you had in mind. I’ll add some more unit tests but I don’t think I’ll have a lot so that’s why I only made a single file

palmskog: Looks good to me, thanks!

examples/compute/src/shared/cv/bir_cv_memScript.sml
```diff
@@ -0,0 +1,209 @@
...
+      case  (bitstring_split (size_of_bir_immtype vty) (b2v result)) of
+        | NONE => NONE
+        | SOME vs =>
+        let vs' = (case en of BEnd_LittleEndian => SOME (REVERSE vs)
+                          |  BEnd_BigEndian => SOME vs
+                          |  BEnd_NoEndian => if (n = 1) then SOME vs else NONE) in
+
+        case vs' of NONE => NONE
+                 |  SOME vs'' => SOME (BCVVal_Mem aty vty (bir_cv_update_mmap aty mmap_alist addr vs''))
+   )
```
palmskog, line 107: @jules-timmerman note that you can do `case ... of` on multiple things at once to simplify the code, e.g., 
```sml
    case (FLOOKUP m1 k, FLOOKUP m2 k) of
    | (SOME v1, SOME v2) => SOME $ f v1 v2
    | (f1, f2) => OPTION_CHOICE f1 f2
```

palmskog (approved):

PR #189, opened 2024-09-13
Fixed RISC-V lifter issue with zero register

didriklundberg:


src/tools/lifter/bir_inst_liftingLib.sml
```diff
@@ -1307,7 +1304,7 @@ fun get_patched_step_hex ms_v hex_code =
   (* Lifting an instruction *)
   (**************************)
 
-  (* Lifting single instructings, is the main workhorse of this library.
+  (* Lifting single instructiions, is the main workhorse of this library.
```
palmskog, line 1307:
```suggestion
  (* Support for lifting single instructions constitutes the main workhorse of this library.
```
Fix typo and reformulate.

src/tools/lifter/selftest_riscv.sml
```diff
@@ -156,6 +156,9 @@ val _ = print_msg "\n";
    * memory address in x2 with offset 8 *)
   val _ = riscv_test_hex_print_asm "SW x14, 8(x2)" "00E12423";
 
+  (* Usage of the zero register: ("SW x0, 0(t2)") *)
+  val _ = riscv_test_hex_print_asm "SW x0, 0(t2)" "0003A023";
```
palmskog, line 160:
```suggestion
  val _ = riscv_test_hex_print_asm "SW x0, 0(x2)" "0003A023";
```
Surely the text should follow conventions with `xN` for general purpose registers?

src/theory/tools/lifter/bir_riscv_extrasScript.sml
```diff
@@ -283,7 +287,7 @@ let
   ];
 
   val elim_zero_thm =
-    GEN_ALL (SIMP_CONV (std_ss++wordsLib.WORD_ss) [] ``riscv_mem_store_byte (a+0w) w mmap``);
+    GEN_ALL (SIMP_CONV (std_ss++wordsLib.WORD_ss) [] ``mem_store_byte (a+0w) w mmap``);
```
palmskog, line 290: I thought the `riscv_mem_` functions were necessary? Is there an obvious reason for why no longer needed here?

palmskog (approved):
Looks good to me, but please take a look at comments.

src/tools/lifter/selftest_riscv.sml
didriklundberg: The rest of the file uses the bare names of the registers so I will change this, however `t2` is the [ABI mnemonic](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-cc.adoc#register-convention) of `x7` (also this can be seen in the selftest log), so your suggestion is wrong. The ASM must have been copy-pasted from somewhere.

src/theory/tools/lifter/bir_riscv_extrasScript.sml
didriklundberg: This occurrence makes more sense if you look at what's going on just above: rewriting of the `riscv_mem` functions in terms of `mem_store_byte`. The usage of `riscv_mem_store_byte` was probably an earlier typo.

I did some tests to understand what's going on here, and I think the current usage of `mem_store_byte` should stay.

palmskog (approved):

PR #194, opened 2024-11-22
Small fix to tutorial

didriklundberg:
This fix makes the instructions in the tutorial README possible to follow. Maybe, we should give this a more complete overhaul.


palmskog:
@didriklundberg I think you know my position on this: we should port it to only use `Holmake`/`Holmakefile`

palmskog (approved):
Sure, let's merge this band-aid, but let's overhaul in the future.

PR #196, opened 2024-12-05
Support for CSR instructions and registers in RISC-V lifter and backlifter

didriklundberg:
Note that while the build of this branch fails, it should succeed when merged into `l3-riscv` (due to the dependency on the updated L3 model).


src/tools/lifter/examples/Holmakefile
```diff
@@ -1,4 +1,4 @@
-INCLUDES = 
+INCLUDES = ..
```
palmskog, line 1:
```suggestion
INCLUDES = $(HOLBADIR)/src/tools/lifter
```

palmskog (approved):
Fine by me, please merge into `l3-riscv` if you think it's ready.

PR #204, opened 2025-08-18
Use Poly/ML 5.9.2 and trindemossen-2

didriklundberg:
This PR updates HolBA to the latest releases of Poly/ML and HOL4.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,4 @@
 # Compiled HOL files
-.HOLMK
-.hollogs
-*Theory.dat
-*Theory.sig
-*Theory.sml
-*.ui
-*.uo
 *~
 *.exe
 *-heap


.gitignore
```diff
@@ -9,6 +9,7 @@
 *~
 *.exe
 *-heap
+.hol/
```
palmskog, line 12: According to release notes, all temporary files now live in `.hol`, so I think we should remove `*Theory.*`, `.HOLMK`, etc., from `.gitignore`.

src/extra/holba_auxiliaryScript.sml
```diff
@@ -96,10 +96,9 @@ QED
 
 
 Theorem INDEX_FIND_EQ_NONE:
-  !l i P. (INDEX_FIND i P l = NONE) <=> (~(EXISTS P l))
+  !l i P. INDEX_FIND i P l = NONE <=> ~EXISTS P l
```
palmskog, line 99:
```suggestion
  !l i P. (INDEX_FIND i P l = NONE) <=> ~EXISTS P l
```
The first parenthesis pair here are actually helpful for readability.

src/tools/comp/bir_compositionLib.sml
```diff
@@ -29,15 +29,15 @@ open bir_inst_liftingHelpersLib;
     open HolBACoreSimps;
   in
 
-    (* These functions obtains the parameters of a
+    (* These functions obtain the parameters of a
      * bir_exec_to_labels_triple *)
```
palmskog, line 33:
```suggestion
    (* These functions obtain the parameters of a bir_exec_to_labels_triple *)
```
Why are we even line-breaking here?

src/tools/comp/bir_compositionLib.sml
```diff
@@ -29,15 +29,15 @@ open bir_inst_liftingHelpersLib;
...
     fun get_contract_prog contract_thm = ((el 1) o snd o strip_comb o concl) contract_thm;
     fun get_contract_l contract_thm = ((el 2) o snd o strip_comb o concl) contract_thm;
     fun get_contract_ls contract_thm = ((el 3) o snd o strip_comb o concl) contract_thm;
     fun get_contract_pre contract_thm = ((el 4) o snd o strip_comb o concl) contract_thm;
     fun get_contract_post contract_thm = ((el 5) o snd o strip_comb o concl) contract_thm;
 
-    (* These functions obtains the parameters of a
+    (* These functions obtain the parameters of a
      * bir_cont *)
```
palmskog, line 41:
```suggestion
    (* These functions obtain the parameters of a bir_cont *)
```

palmskog (dismissed):
Only minor suggestions here, I think this is ready to be merged.

src/extra/holba_auxiliaryScript.sml
didriklundberg: Not sure I agree, `<=>` only operates on Booleans so there can be no ambiguity regarding the `NONE`.

.gitignore
didriklundberg: Yes, I considered this at first, but it might be a problem for residual temporary files if you perform the upgrade in the same repo. IMO keeping them there is fairly benign, but I can remove them if you think it's cleaner.

